PR #4, opened 2018-10-23
Use "top" for potential division by zero when "keep trying" flag is used.

ccasin:
While this is potentially unsound, it's necessary in practice for many binaries I've been toying with.  We should perhaps investigate extending the abstract representation with a notion of "crashed" program in the future.

### DIFF
--- a/value_set/README.md
+++ b/value_set/README.md
@@ -1,37 +1,46 @@
-This plugin is intended to implement value set analysis for the purpose
-of recovering the full control flow graph from a program with indirect jumps.
-It uses circular linear progressions to abstract over sets of possible values.
+This plugin is intended to implement value set analysis for the purpose of
+recovering the full control flow graph from a program with indirect jumps.  It
+uses circular linear progressions to abstract over sets of possible values.
 
 Build
 --------------
+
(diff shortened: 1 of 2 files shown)


jtpaasch:
Looks good to me. Definitely agree that we should figure out how to represent the crashed program.

ivg:
Shouldn't this actually be bottom instead, as the result of division by zero is unreachable? To my understanding is that if there is a program path that includes a division, then this path is terminated, so when we compute the meet-over-all path solution approximation we can safely join with bottom (which would be basically ignored), since this path doesn't exist. Besides the intuition that will also disable widening of any division operation.

ccasin:
@ivg Upon reflection, I think neither top or bottom is right, but your intuition is close.

We're dividing one abstract value v1 by another v2.  v1 and v2 represent many possible concrete values, and v2 includes zero.

Bottom is the wrong answer because it ignores all the possible divisions that come from non-zero values of v2.  Top is the wrong answer because "all value are possible" is not a good model for "something went wrong".

To ignore the potential division by zero I should instead remove 0 from v2 before computing the division.  I'll take a stab at that, which should get us to the "meet over all paths" solution as you suggest.

PR #22, opened 2019-05-31
Convert value_set and explicit_edge to dune

fortunac:
- Converts the `value_set` and `explicit_edge` from oasis to dune. 
- Updates the READMEs to reflect the changes
- Adds `.gitignore`s to the `wp/resources/sample_binaries` directory


ccasin:
I built this and it worked.  A few questions/suggestions:

Right now, the build directions and dependencies are partially documented in the top-level `README.md`, and then there are more detailed instructions in `value_set/README.md`.  Suggest just deleting the stuff in the top-level one and replacing with a pointer to the other, to eliminate any risk of them getting out of sync.

Does the user actually need to install the dependencies in the README before doing the make?  I had forgotten to install `core` on this switch, and opam or dune noticed that and installed it when building `value_set`.  I'd suggest removing any instructions the user doesn't need to explicitly carry out manually.  Although, I had also forgotten to install oUnit, and that wasn't caught until I tried to run the tests, so I'm not totally clear on how the build system is tracking these dependencies.

jtpaasch:
For what it's worth, sometimes depext doesn't install all dependencies. Dustin and I found that it wouldn't install `ppx_deriving`, and so we had to add a line to the README indicating that the user should install that particular package explicitly.

explicit_edge/Makefile
```diff
@@ -1,55 +1,20 @@
...
+.PHONY: all build install clean
 
-doc: setup.ml setup.data build
-	$(SETUP) -doc $(DOCFLAGS)
+all: install
 
-test: setup.ml setup.data build
-	$(SETUP) -test $(TESTFLAGS)
+build:
+	bapbuild -pkg 'cbat_value_set' -tag 'warn(A-48-44)' $(PASS_NAME).plugin
```
codyroux: I recommend adding `-use-ocamlfind` here, which seems to be necessary when using an Ocaml version around `4.05`.

codyroux:
Some slight build shenanigans on my side which can be resolved, though I should probably upgrade to `4.07` ASAP.

PR #39, opened 2019-06-28
Fortunac/increment stack pointer

fortunac:


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -155,12 +155,137 @@ let bv_to_bool (bv : Constr.z3_expr) (ctx : Z3.context) (width : int) : Constr.z
...
+            if Var.equal v rsp then begin
+              warning "RSP is being decremented by %s" (Word.to_string w) ;
+              rhs :: offsets
+            end else offsets
+          | _ -> offsets
+        end else
+          offsets)
+
+let increment_stack_ptr (post : Constr.t) (env : Env.t) (offsets : Exp.t list)
+  : Constr.t =
```
codyroux: Probably not a big deal, but you could pass the `Env.t` here, instead of throwing it away. Here the `Env.t` doesn't get updtated, so it doesn't change, but it would be more uniform.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -155,12 +155,137 @@ let bv_to_bool (bv : Constr.z3_expr) (ctx : Z3.context) (width : int) : Constr.z
...
+  let rsp, _ = Env.get_var env (Var.create "RSP" reg64_t) in
+  List.fold offsets ~init:post ~f:(fun p off ->
+      let z3_off, _, _, _ = exp_to_z3 off env in
```
codyroux: You'd keep the `env` argument here, and return it along with the result.

PR #40, opened 2019-07-01
Add architecture to environment

fortunac:


wp/lib/bap_wp/src/precondition.mli
```diff
@@ -99,8 +99,9 @@ val word_to_z3 : Z3.context -> Bap.Std.Word.t -> Constr.z3_expr
     Returns also the assumptions and the VCs generated by the hooks in the environment. *)
 val exp_to_z3 : Bap.Std.exp -> Env.t -> Constr.z3_expr * Constr.t list * Constr.t list * Env.t
 
-(** The default spec used when mapping subroutines to their preconditions. *)
-val spec_default : Bap.Std.Sub.t -> Env.fun_spec
+(** The default spec used when mapping subroutines to their preconditions.
```
codyroux: I feel like we could describe a bit more what this default spec is, e.g. `the default spec specifies a random output on the output register, if it is modified by the body of the subroutine, and the value of the stack pointer on the return, using a simple heuristic based on the increments of [SP] in the last block before the return`

wp/plugin/wp.ml
```diff
@@ -21,11 +21,16 @@ module Pre = Precondition
 module Env = Environment
 module Constr = Constraint
 
-exception Missing_function of string
+let missing_func_msg (func : string) : string =
+  Format.sprintf "Missing function: %s is not in binary." func
+
+let diff_arch_msg (arch1 : Arch.t) (arch2 : Arch.t) : string =
```
codyroux: Nice!

codyroux:
Looks really good! Maybe worth clarifying what the default spec is doing, though I'm guilty of that as well...

PR #41, opened 2019-07-02
Refactor unit tests into different modules

fortunac:


wp/lib/bap_wp/Makefile
```diff
@@ -14,7 +14,7 @@ install.local: build
 	dune install
 	opam pin add -y .
 
-test: test.unit
+test: test.unit test.performance
```
codyroux: Didn't we remove this for a reason?

wp/lib/bap_wp/tests/performance/test_precondition.ml
```diff
@@ -28,29 +28,24 @@ module BV = Z3.BitVector
...
   let end_time = Sys.time () in
   assert_equal ~ctxt:test_ctx
     ~printer:Z3.Solver.string_of_status
     ~pp_diff:(fun ff (exp, real) ->
-        Format.fprintf ff "\n\n%s\n%!" formatter;
+        Format.fprintf ff "\n\nPost:\n%a\n\nAnalyzing:\n%sPre:\n%a\n\n%!"
+          Constr.pp_constr post (Sub.to_string body) Constr.pp_constr pre;
         print_z3_model ff solver exp real)
     expected result;
   end_time -. start_time
```
codyroux: Maybe we should fail if `end_time -. start_time` is greater than some user-defined threshold: that's the point of performance tests, after all...

wp/lib/bap_wp/Makefile
fortunac: We removed it because the slowdown was coming from building the string rather than Z3. I fixed that, so that isn't the case anymore.

I put it back because we had some changes that broke the performance tests, and I think it would be a good idea to keep track of it even if we don't have that sophisticated of a test case.

codyroux:
Seems good, except I'm not sure about performance tests: it still doesn't feel like we're testing anything useful. Also they don't fail if performance regresses!

wp/lib/bap_wp/Makefile
codyroux: Did they break because of performance regression?

wp/lib/bap_wp/tests/performance/test_precondition.ml
fortunac: Whoops, I forgot about that. I'll put that in.

wp/lib/bap_wp/Makefile
codyroux: If not, might as well add a faster version to the "ordinary" tests...

fortunac: No, it was just an API change that caused the performance tests to be unable to compile

fortunac:
I agree, the performance tests still don't test anything useful. I was planning on making another PR with an updated performance test in the future.

codyroux:
Looks great!

PR #43, opened 2019-07-03
Print out lhs and rhs of an equality constraint

fortunac:
Refuted goals are built as a string in the form:

```
Refuted goals:
EAX0 = EAX09:
        Concrete values: #x80000004 = #x7ffffffd
        Z3 Expression: (let ((a!1 (bvadd #xfffffffd
                  (concat ((_ extract 31 8) (bvadd #xfffffffc EAX09))
                          (bvadd #xfc ((_ extract 7 0) EAX09))))))
                (= EAX0 a!1))
```


codyroux:
Great work!

PR #46, opened 2019-07-23
Remove backedge after call to exit

fortunac:


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -650,9 +654,47 @@ let visit_graph (env : Env.t) (post : Constr.t)
...
+        |> Blk.elts ~rev:true
+      in
+      let exits = Seq.exists elts ~f:(function
+          | `Jmp j -> begin
+              match Jmp.kind j with
+              | Call c -> begin
+                  match Call.target c with
+                  | Direct tid -> begin
+                      match Env.get_sub_name env tid with
+                      | Some target -> String.equal target "exit"
```
codyroux: Probably useful to have this as a parameter to the `filter` function...

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -650,9 +654,47 @@ let visit_graph (env : Env.t) (post : Constr.t)
...
+                      | None -> false
+                    end
+                  | _ -> false
+                end
+              | _ -> false
+            end
+          | _ -> false)
+      in
+      if exits then begin
+        info "Removing the back edge from the return from exit";
```
codyroux: Also would require an update here.

wp/lib/bap_wp/tests/unit/test_precondition.ml
```diff
@@ -911,7 +911,7 @@ let test_loop_3 (test_ctx : test_ctxt) : unit =
              |> Constr.mk_constr
   in
   let pre, _ = Pre.visit_sub env post sub in
-  assert_z3_result test_ctx ctx (Sub.to_string sub) post pre Z3.Solver.SATISFIABLE
+  assert_z3_result test_ctx ctx (Sub.to_string sub) post pre Z3.Solver.UNSATISFIABLE
```
codyroux: Do we have a test that fails if we unfold enough?

codyroux:
All right good job!

PR #50, opened 2019-07-24
Add trivial postcondition to environment in loop_handler

fortunac:


wp/lib/bap_wp/src/environment.ml
```diff
@@ -119,7 +122,7 @@ let wp_rec_call :
   (t -> Constr.t -> start:Graphs.Ir.Node.t -> Graphs.Ir.t -> t) ref =
   ref (fun _ _ ~start:_ _ -> assert false)
 
-let init_loop_unfold (num_unroll : int) : loop_handler =
+let init_loop_unfold (num_unroll : int) (ctx : Z3.context) : loop_handler =
```
codyroux: Not crazy about this `ctx` here, it seems unnecessary and hard to keep consistent with the `ctx` in `env`. Can't we just use the one in `env`?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -531,12 +531,8 @@ let visit_jmp (env : Env.t) (post : Constr.t) (jmp : Jmp.t) : Constr.t * Env.t =
               | Some pre -> pre
               (* We always hit this point when finish a loop unrolling *)
               | None ->
-                (* FIXME: We are passing a trivial condition in the case that this
-                   is a backedge in the middle of the program. However, this
-                   generates more UNSAT results compared to passing in the
-                   postcondition. *)
-                info "Precondition for node %s not found!" (Tid.to_string tid);
-                Constr.mk_goal "true" (Bool.mk_true ctx) |> Constr.mk_constr
+                failwith (Format.sprintf "Precondition for node %s not found!"
```
codyroux: Shouldn't this be an `error` instead of an `info` now?

wp/lib/bap_wp/src/environment.ml
fortunac: I totally forgot about the `ctx` in the `env`. Thanks for the catch.

wp/lib/bap_wp/src/precondition.ml
fortunac: I'm a bit confused here. I removed the `info` made made it a `failwith` with the same message.

codyroux: I mean using `Self().error` to play nicer with BAP logging.

PR #51, opened 2019-07-30
Add flag to check to look for a countermodel

fortunac:


jtpaasch:
Much better comment on that function now.

jtpaasch:
I'd approve it, but I think @codyroux should have a looksie. Any tests needed?

codyroux:
Quick work! We'll add tests for the jump reachability spec in a bit, probably should be a separate PR. I don't think we need to test this, really (famous last words).

PR #55, opened 2019-08-16
Update design decisions with loop unfolding

fortunac:


wp/plugin/design_decisions.md
```diff
@@ -16,3 +16,9 @@
 - We do not correctly handle calls: no effort is made to correctly
   update the variables modified by the call, in particular we assume
   nothing was modified.
+
```
codyroux: I'm not in love with this explanation, since it hides some of the clever thinky things we did to come up with the algorithm, and might be hard to understand even for experts. Why don't I take a stab at it, and you can correct me if I made a mistake?

codyroux:
Closing for now, in favor of an updated version.

PR #56, opened 2019-08-16
Fortunac/output register comparison

fortunac:
Allows the user to specify what registers to check for equivalence with:
`--wp-output-vars=RDI,RSI,RCX,...`

### DIFF
--- a/wp/lib/bap_wp/src/environment.ml
+++ b/wp/lib/bap_wp/src/environment.ml
@@ -166,7 +166,9 @@ let loop_exit_pre (env : t) (node : Graphs.Ir.Node.t) (graph : Graphs.Ir.t)
         ) in
       match exit_node with
       | None -> None
-      | Some e -> get_precondition env e
+      | Some e ->
+        debug "Using precondition from node %s%!" (Tid.to_string e);
+        get_precondition env e
 
 let init_loop_unfold (num_unroll : int) : loop_handler =
   {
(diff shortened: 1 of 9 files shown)


wp/lib/bap_wp/src/environment.ml
```diff
@@ -166,7 +166,9 @@ let loop_exit_pre (env : t) (node : Graphs.Ir.Node.t) (graph : Graphs.Ir.t)
         ) in
       match exit_node with
       | None -> None
-      | Some e -> get_precondition env e
+      | Some e ->
```
codyroux: This seems unrelated to the feature in this PR, maybe we should revert this?

wp/lib/bap_wp/tests/unit/test_compare.ml
```diff
@@ -35,7 +35,7 @@ let assert_z3_compare (test_ctx : test_ctxt) (z3_ctx : Z3.context) (body1 : stri
     ~pp_diff:(fun ff (exp, real) ->
         Format.fprintf ff "\n\nComparing:\n%s\nand\n\n%s\nCompare_prop:\n%a\n\n%!"
           body1 body2 Constr.pp_constr pre;
-        print_z3_model ff solver exp real)
+        print_z3_model ff solver exp real z3_ctx pre)
```
codyroux: Same comment as above.

codyroux:
Other than the minor comments, it looks good!

PR #61, opened 2019-08-22
Compare Z3 arrays by checking the formulas with the Z3 solver

fortunac:
One caveat is that I had to change the API for `get_refuted_goals` from  `t -> Z3.Model.model -> Z3.context -> goal Core_kernel.Sequence.t` to ` t -> Z3.Solver.solver -> Z3.context -> goal Core_kernel.Sequence.t`

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -116,30 +116,47 @@ let substitute (constr : t) (olds : z3_expr list) (news : z3_expr list) : t =
 let substitute_one (constr : t) (old_exp : z3_expr) (new_exp : z3_expr) : t =
   Subst (constr, [old_exp], [new_exp])
 
-let get_refuted_goals_and_paths (constr : t) (model : Z3.Model.model) (ctx : Z3.context)
-  : (goal * path) seq =
+let get_refuted_goals_and_paths (constr : t) (solver : Z3.Solver.solver)
+    (ctx : Z3.context) : (goal * path) seq =
+  let model = Z3.Solver.get_model solver
+              |> Option.value_exn ?here:None ?error:None ?message:None in
   let rec worker (constr : t) (current_path : path) (olds : z3_expr list)
       (news : z3_expr list) : (goal * path) seq =
     match constr with
     | Goal g ->
       let goal_val = Expr.substitute g.goal_val olds news in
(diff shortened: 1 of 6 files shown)


wp/lib/bap_wp/src/constraint.ml
```diff
@@ -116,30 +116,47 @@ let substitute (constr : t) (olds : z3_expr list) (news : z3_expr list) : t =
...
+        | Z3.Solver.SATISFIABLE ->
+          worker c1 (Tid.Map.set current_path ~key:tid ~data:true) olds news
+        | Z3.Solver.UNSATISFIABLE ->
+          worker c2 (Tid.Map.set current_path ~key:tid ~data:false) olds news
+        | Z3.Solver.UNKNOWN ->
+          failwith (Format.sprintf "get_refuted_goals: Unable to resolve branch \
+                                    condition at %s" (Tid.to_string tid))
+      end
     | Clause (hyps, concs) ->
       let hyps_false = List.exists hyps
```
codyroux: Can't you pass *all* the `hyp_val`s in a big list to `Z3.Solver.check`, to only invoke it once? I think this is equivalent to the conjunction, which I think is what you want.

wp/lib/bap_wp/tests/unit/test_compare.ml
```diff
@@ -425,6 +425,32 @@ let test_sub_pair_fun_4 (test_ctx : test_ctxt) : unit =
...
 
+let test_sub_pair_mem_1 (test_ctx : test_ctxt) : unit =
+  let ctx = Env.mk_ctx () in
+  let var_gen = Env.mk_var_gen () in
+  let blk1 = Blk.create () in
+  let blk2 = Blk.create () in
+  let mem = Var.create "mem" (mem32_t `r32) in
+  let loc1 = Var.create "loc" reg32_t in
+  let bv1 = Word.of_int ~width:32 0xDEADBEEF in
+  let bv2 = Word.of_int ~width:32 0x0FA1AFE1 in
```
codyroux: Cute!

codyroux:
Looks fine to me except maybe that one line where you handle `Clause`s! Check my reasoning though on that one...

wp/lib/bap_wp/src/constraint.ml
fortunac: Passing in the list of `hyp_val`s works like a charm!

PR #64, opened 2019-08-28
Inlining now takes a list of subroutines to inline, and uses summarie…

codyroux:
…s for the others


codyroux:
Fixed! Feel free to merge if you want.

codyroux:
Ah dang.

codyroux:
Ok fixed.

Issue #33, opened 2019-06-25
Add preconditions which bound possible values of RSP


codyroux:
This is very X86/X86_64 specific, but I think it's a good idea to have some lower bounds on possible values of `RSP`, just for sanity's sake, say `0xFFFFFFFF00000000 <= RSP <= 0xFFFFFFFFFFFFFFFF - 0x400`.

Not sure it'll help all that much, but it'll be reassuring to have a bit more realistic stack values.

fortunac:
Handled in #107

PR #66, opened 2019-09-13
Create function call predicates

fortunac:
At function call time, the precondition is computed with `Func_pred(in, out) => post[fresh_out/out]`.
After both subroutines have been visited,

`forall in_old, in_new, out_old, out_new. Func_pred(in_old, out_old) ^ Func_pred(in_new, out_new) => in_old = in_new ^ out_old = out_new`

is added to the hypotheses.

Assumptions made:
- both function calls have the same number of input and output variables
- the order of the arguments remains consistent between the two binaries

Lastly, I filtered out `mem` from the variable comparison because it caused Z3 to choke.

### DIFF
--- a/wp/lib/bap_wp/src/compare.ml
+++ b/wp/lib/bap_wp/src/compare.ml
@@ -62,12 +62,12 @@ let fun_call_eqs (env1 : Env.t) (env2 : Env.t) : Constr.t list =
           assert (FuncDecl.equal func_decl1 func_decl2);
           let args1 = Expr.get_args pred1 in
           let args2 = Expr.get_args pred2 in
-          let out_vars = List.append args1 args2 in
-          let sorts = List.map out_vars ~f:Expr.get_sort in
-          let symbols = List.map out_vars
+          let args = List.append args1 args2 in
+          let sorts = List.map args ~f:Expr.get_sort in
+          let symbols = List.map args
               ~f:(fun v -> v |> Expr.to_string |> Z3.Symbol.mk_string ctx) in
-          (* FIXME: Currently assuming that both function calls have the same about
-             of output variables. *)
+          (* FIXME: Currently assuming that both function calls have the same number
+             of input and output variables. *)
           let out_eqs = List.fold2_exn args1 args2 ~init:[]
               ~f:(fun eq var1 var2 -> Bool.mk_eq ctx var1 var2 :: eq) in
           let body = Bool.mk_implies ctx
@@ -78,7 +78,7 @@ let fun_call_eqs (env1 : Env.t) (env2 : Env.t) : Constr.t list =
             |> Constr.mk_goal (Format.sprintf "%s called with equal outputs" f)
             |> Constr.mk_constr
           in
-          f_modified_same_output::hyp
+          f_modified_same_output :: hyp
       )
 
 let compare_blocks
--- a/wp/lib/bap_wp/src/compare.ml
+++ b/wp/lib/bap_wp/src/compare.ml
@@ -18,6 +18,8 @@ include Self()
 
 module Expr = Z3.Expr
 module Bool = Z3.Boolean
+module FuncDecl = Z3.FuncDecl
+module Quant = Z3.Quantifier
 module Env = Environment
 module Pre = Precondition
 module Constr = Constraint
@@ -29,7 +31,7 @@ let set_to_eqs (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t) : Constr.t list 
   (* We keep only the physical variables, since the virtual ones may
      have arbitrary values unrelated to the concrete exectution
      (and in particular be of different types!)
-   *)
+  *)
   let vars = Set.filter vars ~f:(fun v -> Var.is_physical v) in
   Var.Set.fold vars ~init:([], env1, env2)
     ~f:(fun (eqs, env1, env2) v ->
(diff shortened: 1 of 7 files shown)


wp/lib/bap_wp/src/environment.mli
```diff
@@ -114,6 +114,10 @@ val wp_rec_call :
     typically a constant. *)
 val add_var : t -> Bap.Std.Var.t -> Constr.z3_expr -> t
 
+(** Add a function predicate to the environment representing a function call to
+    a target with a given tid. *)
+val add_fun_pred : t -> Bap.Std.Tid.t -> Constr.z3_expr -> t
```
codyroux: So the arguments are fixed by the shape of the function call? Maybe we should elaborate a bit in the comments here.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -314,6 +315,50 @@ let increment_stack_ptr (post : Constr.t) (env : Env.t) (offset : Exp.t)
   let z3_off, _, _, env = exp_to_z3 offset env in
   Constr.substitute_one post sp z3_off, env
 
+let get_vars (t : Sub.t) : Var.Set.t =
```
codyroux: It disturbs me that neither we, nor BAP has this function already written.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -314,6 +315,50 @@ let increment_stack_ptr (post : Constr.t) (env : Env.t) (offset : Exp.t)
...
+
+let mk_fun_pred (env : Env.t) (sub : Sub.t) (out_vars : Constr.z3_expr list)
+  : Constr.t * Env.t =
+  let ctx = Env.get_context env in
+  let fun_name = (Sub.name sub) ^ "_pred" in
+  (* FIXME:
+     1. The order of the arguments is probably not guaranteed to be the same
+        across the two different binaries.
+     2. We are currently filtering out mem from the input variables. We might want
+        to add it back, but it causes a great slowdown when Z3 checks the precondition. *)
```
codyroux: `mem` filtering seems very reasonable to me. At best we might be able to point to *specific* regions of `mem` as being arguments, but not the whole thing.

wp/lib/bap_wp/tests/unit/test_compare.ml
```diff
@@ -425,6 +425,43 @@ let test_sub_pair_fun_4 (test_ctx : test_ctxt) : unit =
     compare_prop Z3.Solver.SATISFIABLE
```
codyroux: Why can't we use BIL to BIR here?

wp/lib/bap_wp/src/compare.ml
```diff
@@ -42,6 +44,43 @@ let set_to_eqs (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t) : Constr.t list
         in (eq::eqs, env1, env2)
       )
 
+(* This hypothesis needs to be generated after visiting both of the subs because
```
codyroux: Maybe a quick note about what it's actually computing might be useful.

wp/lib/bap_wp/src/environment.mli
```diff
@@ -170,6 +174,10 @@ val get_loop_handler :
 (** Obtains the architecture of the program. *)
 val get_arch : t -> Bap.Std.Arch.t
 
+(** Looks up the function predicate of a subroutine in the form of [func_pred(in, out)]
+    with a given Tid that contains information about its input and output variables. *)
```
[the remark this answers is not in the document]
fortunac: The `tid` of the target function. I'll add it to the doc.

wp/lib/bap_wp/tests/unit/test_compare.ml
fortunac: Because I didn't think. Using Bil to Bir will definitely be cleaner here.

wp/lib/bap_wp/src/precondition.ml
fortunac: How do we determine which regions of `mem` is used as as an argument? I can imagine we can determine what arguments are used from the stack based on what is being pushed at call time, and then gets popped base off of instructions like `#33 := mem[RSP, el]:u64`, but I'm not sure how we determine arguments from the heap.

wp/lib/bap_wp/src/environment.mli
fortunac: I'm not sure I understand which arguments you are referring to. Are you referring to the `z3_expr` argument that takes in a predicate of the form `foo(in, ..., out, ...)`?

wp/lib/bap_wp/src/compare.ml
fortunac: Can do.

wp/lib/bap_wp/src/environment.mli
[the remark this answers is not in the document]
fortunac: I just realized that the way I implemented `fun_pred` creation assumes that the same registers will be used as input no matter the call site.

wp/lib/bap_wp/src/environment.mli
codyroux: I was actually referring to the `tid` input; generally, describing what all the non-trivial arguments are supposed to be is a good idea.

wp/lib/bap_wp/src/precondition.ml
codyroux: Best left until future work.

wp/lib/bap_wp/src/environment.mli
codyroux: So the `tid` is a function call site, and the `z3_expr` is the predicate that represents that function call?

fortunac: Updated the doc with more info about the `tid` and `z3_expr`.

codyroux:
We'll revisit this in a bit.

PR #70, opened 2019-09-25
Pzucker/gdb

philzook58:
Added functionality and gdb-filename argument to output a gdb file that will breakpoint at the examined function and fill registers with the counter model.
Added documentation in wp-readme

No test as of yet. Not exactly sure how to unit test this.


wp/lib/bap_wp/src/postprocess.ml
```diff
@@ -0,0 +1,133 @@
+(***************************************************************************)
```
codyroux: Not crazy about the name for this module. Maybe "output.ml"?

codyroux: Or even "output_gdb.ml", I'm not actually sure how good an idea it is to reuse this module for other things.

wp/lib/bap_wp/src/postprocess.ml
```diff
@@ -0,0 +1,133 @@
...
+  | Some v -> v
+  | None -> raise (NoValue (no_value_msg e m))
+
+let word_of_z3_bv bv =
+  let bits = Z3.BitVector.get_size (Z3.Expr.get_sort bv) in
+  let sign = Unsigned in
+  let s = Z3.Expr.to_string bv in
+  word_of_str (String.substr_replace_first s ~pattern:"#" ~with_:"0") bits sign
+
+(* get var will actually create a new variable if it can't find the one we're asking for. *)
```
codyroux: What is `get var`? Is it referring to a `get_var` function?

wp/lib/bap_wp/src/postprocess.ml
```diff
@@ -0,0 +1,133 @@
...
+let reg_names m =
+  m |>
+  Z3.Model.get_decls |>
+  List.map ~f:Z3.FuncDecl.get_name |>
+  List.map ~f:Z3.Symbol.to_string |>
+  List.filter ~f:(fun s -> String.is_substring_at s 0 "R") |>
+  List.map ~f:strip_final_digits
+
+
+(** [output_gdb] is similar to [print_result] except chews on the model and outputs a gdb script with a
```
codyroux: This comment should probably be right before `output_gdb`.

wp/lib/bap_wp/src/postprocess.mli
```diff
@@ -0,0 +1,40 @@
...
+module Env = Environment
+
+module Constr = Constraint
+
+(** Prints out the result from check, and if the result is [SAT], generate a model that
+    represents the registers and memory values that lead to a specific program state. *)
+val print_result : Z3.Solver.solver -> Z3.Solver.status -> Constr.t -> Z3.context -> unit
+
+
+(** Prints to file a gdb script that will fill the appropriate registers with the countermodel *)
```
codyroux: This should be more detailed! E.g. the BAP `mli`s have whole paragraphs explaining the input-output of their functions. We don't need to be quite as detailed, but still. Also: why have a `Z3.context` when there's one in the `Env.t`. Are they the same one? What is that first `string` argument?

wp/plugin/wp.ml
```diff
@@ -148,20 +148,27 @@ let main (file1 : string) (file2 : string)
...
       compare_projs proj file1 file2 var_gen ctx ~func ~check_calls ~inline ~to_inline ~output_vars
     else
       analyze_proj proj var_gen ctx ~func ~inline ~to_inline ~post_cond
   in
   let result = Pre.check solver ctx pre in
-  Pre.print_result solver result pre ctx
+  let () = match gdb_filename with
+          | None -> ()
+          | Some f -> 
+                Printf.printf "DUMPING GDB\n";
```
codyroux: Probably useful to print the filename at this point as well.

wp/plugin/README.md
```diff
@@ -329,6 +329,9 @@ The various options are:
 - `--wp-output-vars=var_list`. List of output variables for equivalence checking
   by `,` given the same input variables in the case of a comparative analysis.
   Defaults to `RAX,EAX` which are the 64- and 32-bit output registers for x86.
+- `--wp-gdb-filename=my_exec.gdb` output a gdb script to file `my_exec.gdb` which 
+  will fill in the . From within gdb, run `source my_exec.gdb` to set a breakpoint at the function given by `--wp-function` and fill the appropriate registers with a found counter-model.
```
fortunac: Can you explain what `.` is?

wp/lib/bap_wp/src/postprocess.ml
```diff
@@ -0,0 +1,133 @@
...
+
+
+(** Implements {!Model}. *)
+
+open Core_kernel
+open Bap.Std
+
+type sign = Unsigned | Signed
+
+let string_of_sign s =
```
fortunac: Can you annotate the types for the inputs and output for the functions in this module?

codyroux:
Oops, looks like this needs a rebase as well to fix merge conflicts.

wp/lib/bap_wp/src/postprocess.ml
philzook58: I should trash a lot of these anyway. They aren't used.

philzook58:
Why does it need a rebase?

wp/plugin/README.md
philzook58: Accidental fragmented sentence. I moved the end of that sentence to the end of the next sentence

wp/lib/bap_wp/src/postprocess.ml
philzook58: Accidental personal notes left in. It was in reference to behavior in the Env module. Removed.

wp/lib/bap_wp/src/postprocess.ml
philzook58: I did also move the print_result function over into here. If I'm going to name it output_gdb.ml I should move that back. output.ml is perfectly acceptable

wp/lib/bap_wp/src/postprocess.ml
philzook58: My statement was false. They are used.

philzook58:
I think it is able to be rebased now.

wp/lib/bap_wp/src/output.ml
```diff
@@ -0,0 +1,126 @@
...
+
+let output_gdb (solver : Z3.Solver.solver) (status : Z3.Solver.status)
+               (env : Env.t) ~func:(func : string) ~filename:(gdb_filename : string) : unit =
+    match status with
+    | Z3.Solver.SATISFIABLE ->
+        let model = Z3.Solver.get_model solver
+                |> Option.value_exn ?here:None ?error:None ?message:None in
+        let regs = reg_names model in
+        Out_channel.with_file gdb_filename  ~f:(fun t -> 
+            Printf.fprintf t "break *%s\n" func; (* The "*" is necessary to break before some slight setup *)
```
fortunac: Not super important, but I noticed we use both the `Printf` modules and the `Format` modules. What's the difference between these two and would it be a good idea to be uniform?

philzook58: I have not figured out how to get Format to actually print to a file. Me and jt scratched our heads about this one and couldn't figure it out. 

```ocaml
        Out_channel.with_file gdb_filename  ~f:(fun t -> 
            let t = Format.formatter_of_out_channel t in
            Format.fprintf t "break *%s\n" func; (* The "*" is necessary to break before some slight setup *)
            Format.fprintf t "start\n";
            List.iter regs ~f:(fun r -> 
                  let w = val_for_reg r model env in
                  let s = Format.asprintf "set $%s = %a \n" (String.lowercase r) Bap.Std.Word.pp_hex w  in
                  Format.fprintf t "%s" s))   
```

outputs a blank file. Thoughts? Printf was a workaround.

philzook58: Tried some various flush commands too.

fortunac: That's super interesting. Flushing sounds like the intuitive answer to this. If that's the case, keeping it as `Printf` is all good with me.

PR #71, opened 2019-09-26
At a function call, generate a z3_expr that represents function outputs

fortunac:


wp/lib/bap_wp/src/environment.ml
```diff
@@ -36,6 +37,7 @@ type t = {
   precond_map : Constr.t TidMap.t;
   fun_name_tid : Tid.t StringMap.t;
   call_map : string TidMap.t;
+  fun_outputs_map : FuncDecl.func_decl list TidMap.t;
```
codyroux: Now I wonder if we should have comments explaining all these fields...

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -314,6 +315,60 @@ let increment_stack_ptr (post : Constr.t) (env : Env.t) (offset : Exp.t)
...
+  sub
+  |> Sub.ssa
+  |> Sub.free_vars
+  |> Var.Set.filter ~f:(fun v -> not (Target.CPU.is_mem v))
+  |> Var.Set.to_list
+
+(* Creates a Z3 function of the form func_ret_out_var(in_vars, ...) which represents
+   an output variable to a function call. It is added to the environment, and
+   substitutes the original z3_expr representing the output variable. *)
+let subst_fun_outputs (env : Env.t) (sub : Sub.t) (post : Constr.t)
```
codyroux: If you're building these here each time, why do they need to be in the `Env`?

fortunac: You're right. We don't really need them in the `env`. It would probably be better to remove that field.

wp/lib/bap_wp/src/environment.ml
fortunac: Gonna remove this field from the `env`.

codyroux:
Looks good to me. @philzook58 you want to take a gander?

philzook58:
Looks good to me from what I understand of it. Compiles and runs on my machine. Tests pass.

Issue #74, opened 2019-10-01
Preserving names of registers 


fortunac:
In `strip_final_digits`:

https://github.com/draperlaboratory/cbat_tools/blob/7a8d7c6c42b9b0011fdbb38008c769655ec1e1f5/wp/lib/bap_wp/src/output.ml#L60

We are also stripping out the 8-12 from registers r8 - r12, and as a result we are setting a nonexistent registers, i.e.:

```
set $r = 0
```

philzook58:
The commit above seems to have addressed this issue

PR #76, opened 2019-10-02
Changed Output gdb to use bap registers.

philzook58:
Removed unused functions for the moment. Perhaps make a util.ml file later.
Hopefully should address #74. I have not confirmed so, I don't currently have an example that uses numbered registers, but I have no reason to believe that should happen anymore as we're no longer stripping numbers.


fortunac:
Ran this on our retrowrite example, and it works great!

codyroux:
Looks great, thanks Phil!

PR #78, opened 2019-10-03
Improved model printing

codyroux:
Function comparisons now return pairs of updated environments, in the order that they were passed in.

Printing of counter-models is now significantly nicer, but we only print variables which appear in the `var_map` of the first environment.


wp/plugin/wp.ml
```diff
@@ -106,7 +106,7 @@ let analyze_proj (proj : project) (var_gen : Env.var_gen) (ctx : Z3.context)
   let pre, env' = Pre.visit_sub env post main_sub in
   Format.printf "\nSub:\n%s\nPre:\n%a\n%!"
     (Sub.to_string main_sub) Constr.pp_constr pre;
-  (pre, env')
+  (pre, env', env)
```
fortunac: Can you explain to me why we need both `env'` and `env` here?

PR #81, opened 2019-10-07
Fortunac/print paths

fortunac:
Prints out the path of refuted goal like this:
```
Refuted goals:
assert_fail:
    Z3 Expression: false
    Path:
    00000303: when ZF goto %000002fb (taken) (Address: 0x400433:64u)
```

In the case where BAP cannot find the address of a jump:
```
Refuted goals:
RAX0 = RAX067:
    Concrete values: = #x0000000000000000 #x0000000000000001
    Z3 Expression: = #x0000000000000000 #x0000000000000001
    Path:
    00000878: when ZF goto %00000870 (taken)
    00000881: when ZF goto %00000879 (not taken)
    00000f82: when CF goto %00000f7a (taken)
    00000fa2: goto %00000894 (taken)
    00001170: goto %00000f4b (taken)
```

This has some breaking changes with Primus. `jmp`s should be used instead of `tid`s when creating a jump spec and creating an `ITE` constraint.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -19,7 +19,7 @@ module Bool = Z3.Boolean
 
 type z3_expr = Expr.expr
 
-type path = bool Bap.Std.Tid.Map.t
+type path = bool Jmp.Map.t
 
 type goal = { goal_name : string; goal_val : z3_expr }
 
@@ -56,7 +56,7 @@ let get_goal_val (g : goal) : z3_expr =
 
 type t =
   | Goal of goal
-  | ITE of Tid.t * z3_expr * t * t
+  | ITE of Jmp.t * z3_expr * t * t
   | Clause of t list * t list
   | Subst of t * z3_expr list * z3_expr list
 
@@ -65,7 +65,7 @@ let rec pp_constr (ch : Format.formatter) (constr : t) : unit =
   | Goal g -> Format.fprintf ch "%s" (goal_to_string g)
   | ITE (tid, e, c1, c2) ->
     Format.fprintf ch "ITE(%s, %s, %a, %a)"
-      (Tid.to_string tid) (Expr.to_string e) pp_constr c1 pp_constr c2
+      (tid |> Term.tid |> Tid.to_string) (Expr.to_string e) pp_constr c1 pp_constr c2
   | Clause (hyps, concs) ->
     Format.fprintf ch "(";
     (List.iter hyps ~f:(fun h -> Format.fprintf ch "%a" pp_constr h));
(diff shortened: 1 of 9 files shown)


wp/lib/bap_wp/src/constraint.ml
```diff
@@ -125,7 +143,7 @@ let get_refuted_goals_and_paths (constr : t) (solver : Z3.Solver.solver)
     match constr with
     | Goal g ->
       let goal_val = Expr.substitute g.goal_val olds news in
-      let goal_res = Option.value_exn (Z3.Model.eval model goal_val true) in
+      let goal_res = Option.value_exn (Model.eval model goal_val true) in
```
codyroux: I wonder if we should implement `Option.value_exn (Model.eval m v true)` since we call it so much, and we don't get a useful error if this fails currently...

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -16,34 +16,52 @@ open Bap.Std
...
+  Format.fprintf fmt "%s:" g.goal_name;
   if Bool.is_eq g.goal_val then begin
     let args = Expr.get_args g.goal_val in
-    Buffer.add_string buf "\n\tConcrete values: = ";
+    Format.fprintf fmt "\n\tConcrete values: = ";
     List.iter args ~f:(fun arg ->
-        let value = Option.value_exn (Z3.Model.eval model arg true) in
-        Buffer.add_string buf (Format.sprintf "%s " (Expr.to_string value)));
-    Buffer.add_string buf "\n\tZ3 Expression: = ";
+        let value = Option.value_exn (Model.eval model arg true) in
```
codyroux: See below.

wp/lib/bap_wp/src/constraint.ml
fortunac: Probably a good idea. I just grepped and found at least 7 cases where we called `Option.value_exn (Model.eval m v true)`

wp/lib/bap_wp/tests/unit/test_utils.ml
```diff
@@ -80,14 +80,14 @@ let print_z3_model (ff : Format.formatter) (solver : Z3.Solver.solver)
...
           Format.fprintf ff "%s\n%!" (Constr.refuted_goal_to_string g model))
 
-(* Obtains the tid of a jump in a sub given its jump condition. *)
-let jump_tid (sub : Sub.t) (cond : Exp.t) : Tid.t =
+(* Finds a jump in a subroutine given its jump condition. *)
+let find_jump (sub : Sub.t) (cond : Exp.t) : Jmp.t =
   let jumps =
     Term.enum blk_t sub
-    |> Seq.map ~f:(fun b -> Term.enum jmp_t b)
+    |> Seq.map ~f:(Term.enum jmp_t)
```
codyroux: Nice eta-reduction!

codyroux:
Looks good!

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -44,14 +44,22 @@ let pp_path (ch : Format.formatter) (p : path) : unit =
 let path_to_string (p : path) : string =
   Format.asprintf "%a" pp_path p
 
+let eval_model_exn (model : Model.model) (expr: z3_expr) : z3_expr =
+  Model.eval model expr true
+  |> Option.value_exn
+    ?here:None
+    ~error:(Error.of_string "eval_model_exn: Error evaluating expression with model.")
```
codyroux: Maybe "Error evaluating expression: model not found". Or something...

codyroux:
Looks good!

PR #86, opened 2019-10-17
Fortunac/retrowrite bug fixes

fortunac:
Fixes a bug in `wp`, where we were looking at the target of a function call to determine how much to increment the stack pointer when visiting a call site. The correct behavior is to increment the stack pointer by 8 for the return address. We will have to update this in the future when we start supporting other architectures.

### DIFF
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -247,17 +247,26 @@ let set_fun_called (post : Constr.t) (env : Env.t) (tid : Tid.t) : Constr.t =
                                       Option.value_exn ?here:None ?error:None ?message:None) in
   Constr.substitute_one post fun_name (Bool.mk_true ctx)
 
+let is_x86 (a : Arch.t) : bool =
+  match a with
+  | `x86 | `x86_64 -> true
+  | _ -> false
+
 let increment_stack_ptr (post : Constr.t) (env : Env.t) : Constr.t * Env.t =
-  let ctx = Env.get_context env in
   let arch = Env.get_arch env in
-  let module Target = (val target_of_arch arch) in
-  let sp, env = Env.get_var env Target.CPU.sp in
-  let width = Target.CPU.sp |> Var.typ |> typ_size in
-  (* TODO: We are currently incrementing the stack pointer by 8 for the return address.
-     We will have to change that offset depending on the archtecture. *)
-  let offset = BV.mk_numeral ctx "8" width in
-  let z3_off = BV.mk_add ctx sp offset in
-  Constr.substitute_one post sp z3_off, env
+  if is_x86 arch then
+    begin
+      let module Target = (val target_of_arch arch) in
+      let sp, env = Env.get_var env Target.CPU.sp in
+      let width = Target.CPU.sp |> Var.typ |> typ_size in
+      let addr_size = arch |> Arch.addr_size |> Size.in_bytes in
+      let ctx = Env.get_context env in
+      let offset = BV.mk_numeral ctx (Int.to_string addr_size) width in
+      let z3_off = BV.mk_add ctx sp offset in
+      Constr.substitute_one post sp z3_off, env
+    end
+  else
+    post, env
 
 let get_vars (t : Sub.t) : Var.Set.t =
   let visitor =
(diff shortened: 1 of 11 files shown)


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -232,88 +233,30 @@ let exp_to_z3 (exp : Exp.t) (env : Env.t) : Constr.z3_expr * Constr.t list * Con
...
+let increment_stack_ptr (post : Constr.t) (env : Env.t) : Constr.t * Env.t =
+  let ctx = Env.get_context env in
   let arch = Env.get_arch env in
   let module Target = (val target_of_arch arch) in
-  let sp, _ = Env.get_var env Target.CPU.sp in
-  let z3_off, _, _, env = exp_to_z3 offset env in
+  let sp, env = Env.get_var env Target.CPU.sp in
+  let width = Target.CPU.sp |> Var.typ |> typ_size in
+  (* TODO: We are currently incrementing the stack pointer by 8 for the return address.
+     We will have to change that offset depending on the archtecture. *)
```
ivg: It actually ABI dependent but, fortunately, mostly relative to x86 family. Basically, if you will assume a C ABI, then you should decrement it by 4 for x86 and by 8 for x86_64. For all other architectures supported by BAP (and known to ABIs) there is no need for this correction.

Therefore, I'd suggest the following:
1) increment only if the architecture is `#x86` 
2) increment by 4 or 8 depending on arch (I would suggest: `Size.in_bytes (Arch.addr_size arch)`)

Then it will work fine for all architectures in the nearest future :)

ivg: see also a realtive PR for Primus
https://github.com/BinaryAnalysisPlatform/bap/pull/955/files

codyroux: I see, interesting! There does seem to be some effort duplication across WP and Primus, but it's hard for me to see if we could easily resolve this (probably not).

ivg: The main effort here is basically researching this problem, which is basically contained in the PR message, which is worthwhile to read :) Another commonality is that any analysis requires some kind of stubs and one of the reasons why many people started to use Primus is because how easy it is to stub things with Primus Lisp. That's why we decided to generalize Primus Lisp, and it will now become BAP Lisp. The plan was to release it together with BAP 2.0 but I didn't have time to do this, but will return to it as soon as will get it released (basically in a couple of weeks). We will soon be able to use BAP Lisp to specify facts for BAP's knowledge base. The first example that comes to mind (other than the already existing stubs) is writing lifters. And I believe that WP can also benefit from stubs written in BAP Lisp.  

But back to this PR. There is one caveat, and I can't infer immediately from the PR, whether it is handled correctly. The stack pointer is corrected by the subroutine code (when the `ret` instruction is executed). Therefore if the subroutine code is present, there should be no corrections (otherwise we will overcorrect). So we should be careful and do corrections only for stubs (and for x86).

codyroux: Right, we are aware of the issue, and we'll fix this in a cleaner way at some point. 

For now things happen to work however, because of how we treat inlined code: basically the call becomes a `nop` and the return as well, because of a silly bug (we treat indirect jumps as `nop`s).

fortunac: Thanks for taking a look at this Ivan! I was unaware that the correction was only needed in x86. Right now, we're only correcting the stack pointer when we are using our stubs to handle the targets of the function calls.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -388,17 +340,16 @@ let spec_verifier_error (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
...
-    let offset = get_stack_ptr_offset sub arch in
     let open Env in
     Some {
       spec_name = "spec_verifier_assume";
       spec = Summary
           (fun env post tid ->
              let ctx = Env.get_context env in
              let post = set_fun_called post env tid in
-             let post, env = increment_stack_ptr post env offset in
+             let post, env = increment_stack_ptr post env in
```
codyroux: This line seems to be repeated in every spec. I wonder if we'll be adding anything to this in the future? If so, we should be creating a `function_epilogue` or something, that handles this kind of book-keeping. Probably premature, but something to keep in mind.

codyroux:
This looks really great. Great job, Chloe!

PR #89, opened 2019-10-25
Pzucker/gdb mem

philzook58:
After a long journey of Z3 build troubles and api confusion, adds the ability to extract memory models for gdb-output. Having said that, they are totally faulty and case gdb to fail immediately when they try to write to memory locations it has no access to.

I am concerned about what it will do if it encounters a lambda model, but it is profitable to open a pull review at this point I think.

### DIFF
--- a/wp/lib/bap_wp/src/output.ml
+++ b/wp/lib/bap_wp/src/output.ml
@@ -133,12 +133,11 @@ let output_gdb (solver : Solver.solver) (status : Solver.status)
     Printf.printf "Z3 Version: %s\n" (Z3.Version.to_string);
     let model = Solver.get_model solver
                 |> Option.value_exn ?here:None ?error:None ?message:None in
-    Printf.printf "%s \n" (Z3.Model.to_string model);
     let option_mem_model = get_mem model in
     let varmap = Env.get_var_map env in 
 >>>>>>> 874382e... Tests and simple mem-model extractor for gdb
     let module Target = (val target_of_arch (Env.get_arch env)) in
-    let regmap = varmap in (* VarMap.filter_keys ~f:(Target.CPU.is_reg) varmap in *)
+    let regmap = VarMap.filter_keys ~f:(Target.CPU.is_reg) varmap in 
     let reg_val_map = VarMap.map ~f:(fun z3_reg -> Constr.eval_model_exn model z3_reg) regmap in
     Out_channel.with_file gdb_filename  ~f:(fun t ->
         Printf.fprintf t "break *%s\n" func; (* The "*" is necessary to break before some slight setup *)
--- a/wp/lib/bap_wp/src/output.ml
+++ b/wp/lib/bap_wp/src/output.ml
@@ -66,7 +66,6 @@ type mem_model = {default : Constr.z3_expr ; model : (Constr.z3_expr * Constr.z3
 
 let extract_array (e : Constr.z3_expr) : mem_model  = 
       let rec extract_array' (partial_map : (Constr.z3_expr * Constr.z3_expr) list) (e : Constr.z3_expr) : mem_model =
-          Printf.printf "%s\n" (Z3.Expr.to_string e);
           let numargs = Z3.Expr.get_num_args e in
           let args = Z3.Expr.get_args e in
           let f_decl = Z3.Expr.get_func_decl e in
@@ -83,19 +82,28 @@ let extract_array (e : Constr.z3_expr) : mem_model  =
                 { default = key; model = List.rev partial_map}
                 end
           else begin
-                warning "Unexpected case destructing Z3 array: %s" f_name;
+                warning "Unexpected case destructing Z3 array: %s" (Z3.Expr.to_string e);
                 {default = e ; model = partial_map}
                 end
       in
       extract_array' [] e
 
-let get_mem (m : Z3.Model.model) : mem_model option =
-   let decls = Z3.Model.get_decls m |> List.filter ~f:(fun decl -> (Z3.Symbol.to_string (Z3.FuncDecl.get_name decl)) = "mem0") in
-   match (List.hd decls) with
-   | None -> warning "No mem0 found.";
-             None
-   | Some f_decl -> let f_interp = Option.value_exn (Z3.Model.get_const_interp m f_decl) in
-                    Some (extract_array f_interp)
+let is_x86 (a : Arch.t) : bool =
+  match a with
+  | #Arch.x86 -> true
+  | _ -> false
+
+let get_mem (m : Z3.Model.model) (env : Env.t) : mem_model option =
+    let arch = Env.get_arch env in
+    if is_x86 arch then
+    begin
+      let module Target = (val target_of_arch arch) in
+      let mem, _ = Env.get_var env Target.CPU.mem in
+      Some (extract_array (Constr.eval_model_exn m mem))
+    end
+    else
+      None
+
 
 
 let print_result (solver : Solver.solver) (status : Solver.status)
@@ -130,7 +138,7 @@ let output_gdb (solver : Solver.solver) (status : Solver.status)
     Printf.printf "Z3 Version: %s\n" (Z3.Version.to_string);
     let model = Solver.get_model solver
                 |> Option.value_exn ?here:None ?error:None ?message:None in
-    let option_mem_model = get_mem model in
+    let option_mem_model = get_mem model env in
     let varmap = Env.get_var_map env in 
 >>>>>>> 874382e... Tests and simple mem-model extractor for gdb
     let module Target = (val target_of_arch (Env.get_arch env)) in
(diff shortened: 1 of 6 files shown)


wp/lib/bap_wp/src/output.mli
```diff
@@ -39,3 +39,10 @@ val print_result : Z3.Solver.solver -> Z3.Solver.status -> Constr.t
 
 (** Prints to file a gdb script that will fill the appropriate registers with the countermodel *)
 val output_gdb : Z3.Solver.solver -> Z3.Solver.status -> Env.t -> func:string -> filename:string -> unit
+
+type mem_model = {default : Z3.Expr.expr ; model : (Z3.Expr.expr * Z3.Expr.expr) list}
```
fortunac: Can you add a comment describing what `default` and `model` is supposed to be?

wp/lib/bap_wp/src/output.ml
```diff
@@ -60,6 +60,48 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
   Format.flush_str_formatter ()
 
 
+type mem_model = {default : Z3.Expr.expr ; model : (Z3.Expr.expr * Z3.Expr.expr) list}
```
fortunac: Not super important, but we've been using `Constr.z3_expr` instead of `Z3.Expr.expr` and I think it would be good to be consistent.

wp/lib/bap_wp/src/output.ml
```diff
@@ -60,6 +60,48 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
...
+                end
+          else begin
+                Printf.printf "Unpexpected case destructing Z3 array: %s" f_name;
+                {default = e ; model = partial_map}
+                end
+      in
+      extract_array' [] e
+
+
+module F = Z3.Model.FuncInterp
```
fortunac: Do we need to have this module redefined here if it was defined earlier as `module FInterp = Model.FuncInterp`?

philzook58: I should move this definition up. I don't see the FInterp definition?

philzook58: Oh I'm blind

wp/lib/bap_wp/src/output.ml
```diff
@@ -60,6 +60,48 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
...
+                let key = List.nth_exn args 1 in
+                let value = List.nth_exn args 2 in
+                extract_array' (( key , value ) :: partial_map) next_arr
+                end
+          else if ( Int.(numargs = 1) && String.(f_name = "const")) then begin
+                let key = List.nth_exn args 0 in
+                { default = key; model = List.rev partial_map}
+                end
+          else begin
+                Printf.printf "Unpexpected case destructing Z3 array: %s" f_name;
```
fortunac: Might want to make the a warning statement:
`warning "Unexpected ... %s\n" f_name;`

wp/lib/bap_wp/src/output.ml
philzook58: It actually looks like I don't even use F

wp/lib/bap_wp/tests/unit/test_output.ml
```diff
@@ -0,0 +1,33 @@
+open !Core_kernel
+open OUnit2
+open Bap_wp
+open Output
+
+
+
+let test_get_model (test_ctx : test_ctxt) : unit =
```
fortunac: Nice!!

wp/lib/bap_wp/src/output.ml
```diff
@@ -60,6 +60,48 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
...
+      in
+      extract_array' [] e
+
+
+module F = Z3.Model.FuncInterp
+
+let get_mem (m : Z3.Model.model) : mem_model option =
+   let decls = Z3.Model.get_decls m |> List.filter ~f:(fun decl -> (Z3.Symbol.to_string (Z3.FuncDecl.get_name decl)) = "mem0") in
+   match (List.hd decls) with
+   | None -> Printf.printf "No mem0 found.";
```
fortunac: Also might want to make this a warning.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -895,7 +895,8 @@ let check ?refute:(refute = true) (solver : Z3.Solver.solver) (ctx : Z3.context)
     else
       pre'
   in
-  Z3.Solver.check solver [is_correct]
+  let () = Z3.Solver.add solver [is_correct] in
```
fortunac: What's the difference between running the solver on `is_correct` vs. adding `is_correct` then running the solver?

wp/lib/bap_wp/src/output.ml
fortunac: Might be a good idea to remove the definition if we don't use it

fortunac:
Looks good to me. Great work!

wp/lib/bap_wp/src/precondition.ml
philzook58: This is the change that made the simple compare test time out vs not time out.
Here's Bjorner's answer about why this is that I don't completely understand
https://github.com/Z3Prover/z3/issues/2654

philzook58: I think more simplifications are run with this new form

wp/lib/bap_wp/src/output.ml
```diff
@@ -60,6 +60,45 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
   Format.flush_str_formatter ()
 
 
+type mem_model = {default : Constr.z3_expr ; model : (Constr.z3_expr * Constr.z3_expr) list}
+
+(** [extract_array] takes a z3 expression that is a seqeunce of store and converts it into
```
codyroux: > a seqeunce of store

wp/lib/bap_wp/src/output.ml
```diff
@@ -60,6 +60,45 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
...
+                warning "Unexpected case destructing Z3 array: %s" f_name;
+                {default = e ; model = partial_map}
+                end
+      in
+      extract_array' [] e
+
+let get_mem (m : Z3.Model.model) : mem_model option =
+   let decls = Z3.Model.get_decls m |> List.filter ~f:(fun decl -> (Z3.Symbol.to_string (Z3.FuncDecl.get_name decl)) = "mem0") in
+   match (List.hd decls) with
+   | None -> warning "No mem0 found.";
```
codyroux: I wonder if there's a way to be more agnostic about the name of the `mem` variable, e.g. by traipsing around a `CPU` or `Target` instance to be able to call `CPU.mem` from here: http://binaryanalysisplatform.github.io/bap/api/v1.6.0/Bap.Std.CPU.html

And get the name for the memory variable.

codyroux: See e.g. what we do here: https://github.com/draperlaboratory/cbat_tools/blob/9f1180be9e01c4d51c5ab203494c062d6ef1c291/wp/lib/bap_wp/src/precondition.ml#L260

wp/lib/bap_wp/src/output.ml
```diff
@@ -60,6 +60,45 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
...
+                let key = List.nth_exn args 1 in
+                let value = List.nth_exn args 2 in
+                extract_array' (( key , value ) :: partial_map) next_arr
+                end
+          else if ( Int.(numargs = 1) && String.(f_name = "const")) then begin
+                let key = List.nth_exn args 0 in
+                { default = key; model = List.rev partial_map}
+                end
+          else begin
+                warning "Unexpected case destructing Z3 array: %s" f_name;
```
codyroux: Maybe print the whole expression here? Might be hard to debug otherwise.

wp/lib/bap_wp/tests/unit/test.ml
```diff
@@ -15,9 +15,10 @@ open OUnit2
 
 let suite =
   "Unit Tests" >::: [
-    "Precondition" >::: Test_precondition.suite;
+    "Precondition" >::: Test_precondition.suite;
```
codyroux: Any reason there's an extra space here?

wp/lib/bap_wp/tests/unit/test_output.ml
codyroux: Agreed!

wp/plugin/wp.ml
```diff
@@ -106,7 +106,7 @@ let analyze_proj (proj : project) (var_gen : Env.var_gen) (ctx : Z3.context)
   let pre, env' = Pre.visit_sub env post main_sub in
   Format.printf "\nSub:\n%s\nPre:\n%a\n%!"
     (Sub.to_string main_sub) Constr.pp_constr pre;
-  (pre, env', env)
+  (pre, env, env')
```
codyroux: I'm a bit disturbed by this bug. Is there any way we can return more structured output to avoid these kinds of shenanigans in the future? This is probably a task for myself.

codyroux:
Looks good Phil! My comments are quite minor, as usual. If you want to make fixes, great, otherwise I can rebase as soon as the merge conflicts are resolved.

PR #90, opened 2019-10-29
Fortunac/registers along path

fortunac:
If the plugin is run with the `--wp-print-path` flag, when a model is generated, the path along with the registers/flags at each jump in the path will be printed based off the values in the model. Ex:
```
Path:
        0000063c: when ZF goto %00000634 (taken) (Address: 0x4004F4:64u)
        ZF   |->  #b1
        SF   |->  #b0
        RSP  |->  #x0000000000000008
        RDX  |->  #x0000000000000000
        RDI  |->  #x0000000000000000
        RCX  |->  #x0000000000000000
        RBP  |->  #x0000000000000000
        RAX  |->  #x0000000000000000
        PF   |->  #b1
        OF   |->  #b0
        CF   |->  #b0
        AF   |->  #b0

        0000063c: when ZF goto %00000634 (not taken) (Address: 0x4004F5:64u)
        ZF   |->  #b0
        SF   |->  #b1
        RSP  |->  #xffffffffffffffe8
        RDX  |->  #x0000000000000000
        RCX  |->  #x0000000000000000
        RBP  |->  #xfffffffffffffff8
        RAX  |->  #x0000000000000003
        PF   |->  #b1
        OF   |->  #b0
        CF   |->  #b1
        AF   |->  #b1
```

I had to change the API a bit because I made a `refuted_goal` data type which includes the goal, path, and map of register values.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -159,9 +159,12 @@ let get_refuted_goals (constr : t) (solver : Z3.Solver.solver)
     | ITE (jmp, cond, c1, c2) ->
       let cond_val = Expr.substitute cond olds news in
       let cond_res = eval_model_exn model cond_val in
+      (* FIXME: Still getting some values that don't necessarily match up with
+         their values when running the program in GDB. *)
       let registers =
         List.fold (List.zip_exn olds news) ~init:[]
           ~f:(fun regs (reg, value) ->
+              (* Manually removing mem from the list of variables being updates. *)
               if String.is_prefix ~prefix:"mem" (Expr.to_string reg) then
                 regs
               else
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -136,6 +136,23 @@ let substitute (constr : t) (olds : z3_expr list) (news : z3_expr list) : t =
 let substitute_one (constr : t) (old_exp : z3_expr) (new_exp : z3_expr) : t =
   Subst (constr, [old_exp], [new_exp])
 
+let update_current_regs (model : Model.model) (regs : z3_expr list) (vals : z3_expr list)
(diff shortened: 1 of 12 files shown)


wp/lib/bap_wp/src/constraint.ml
```diff
@@ -23,27 +23,11 @@ type z3_expr = Expr.expr
 
 type path = bool Jmp.Map.t
 
-type goal = { goal_name : string; goal_val : z3_expr }
+type reg_map = (z3_expr * z3_expr) list Jmp.Map.t
 
-let goal_to_string (g : goal) : string =
-  Format.sprintf "%s: %s%!" g.goal_name (Expr.to_string (Expr.simplify g.goal_val None))
+type goal = { goal_name : string; goal_val : z3_expr }
```
philzook58: This isn't a comment on this pull request in particular, but goal seems like a slightly odd name for this concept

philzook58:
Looks Good!

wp/lib/bap_wp/src/output.ml
```diff
@@ -41,86 +35,75 @@ let format_model (model : Model.model) (env1 : Env.t) (_env2 : Env.t) : string =
           (key, value)::pairs)
   in
   let fmt = Format.str_formatter in
-  List.iter key_val
-    ~f:(fun (var, value) ->
-        let var_str = Var.to_string var in
-        let pad_size = Int.max (5 - (String.length var_str)) 1 in
```
codyroux: My beautiful code!

codyroux:
Seems good to me, though I suspect there are opportunities for refactoring (I guess that's always true...).

Good job Chloe!

wp/lib/bap_wp/src/output.ml
fortunac: It's still there! I just moved it to `constraint.ml`, so I could reuse it to print out the registers in the path.

PR #96, opened 2019-11-12
Update sample binary makefiles to prevent bug in saving projects

fortunac:
Fixes a bug we had with saving projects. Adds a target `test.clean` that will clean all the sample binaries.


codyroux:
This is black magic to me, I'll assume it works fine.

PR #97, opened 2019-11-12
Pad binop instructions so that the operands have the same bitwidth.

fortunac:
It seems that it is possible bit shift an address on a 2-bit scale, and in this case, we should pad the smaller operand to have the same bitwidth as the wider operand.

More info here: https://wiki.osdev.org/X86-64_Instruction_Encoding#ModR.2FM_and_SIB_bytes


codyroux:
Nice and clean!

PR #101, opened 2019-11-21
Pzucker/precondition

### LINKED ISSUE #49: Add concrete values as preconditions
Allow the user of the WP to add a number of concrete values for the parts of the input environment, say by giving a set of pairs of expressions and values:
`val visit_sub : Env.t -> Constr.t -> Bap.Std.Sub.t -> Constr.t * Env.t`

becomes:

`val visit_sub : Env.t -> (Exp.t * Val.t) seq -> Constr.t -> Bap.Std.Sub.t -> Constr.t * Env.t`

Or maybe modify `check` to be able to re-run the solver with different possible values.

Alternately just separate this out into a new operation that adds constraints to the pre-condition.

### LINKED ISSUE #100: Models are not being printed out
When `wp` is run on a single program and results in `SAT!`, there is no model printed out to the screen.

It seems there is a Z3 model with the correct values, but we lose the values as we are formatting them.

Some samples not working are `simple_wp` and `return_argc`.

philzook58:
- adds the ability the specify assumptions at the beginning of a program about the variables. Closes #49 ? Does not address idea about soft constraints
- It does a string find and replace on variables in smt-lib expressions given to convert them to the internal mangled names.
- Contains bug fix for popping BIL Let bindings from the Env.var_map
- bug fix for the returned environments from analyze_proj in wp.ml. Should fix #100 
- Moved smt-lib related functions out of precondition.ml
- Adds the ability to create custom preconditions and postconditions in comparative analysis. The variables are referenced by appending "_orig" and "_mod" to variable names.

 
Example usage
```
bap resources/sample_binaries/dummy/hello_world.out --pass=wp --wp-compare=true --wp-file1=resources/sample_binaries/equiv_argc/main_1.bpj --wp-file2=resources/sample_binaries/equiv_argc/main_2.bpj --wp-function=main --wp-output-vars=RAX,EAX --wp-precond="(assert (= RDI_mod #x0000000000000003))  (assert (= RDI_orig #x0000000000000003))"
```

```
bap resources/sample_binaries/simple_wp/main --pass=wp --wp-precond="(assert (= RDI #x0000000000000002))"
```

TODO:
- Does the naming make sense for precondition? Should it be called assumptions or hypotheses?
- I'll also need to rebase off of Chloe's pull request 
- I still need to add tests
- The placement of the new smt-lib code in compare.ml feels a bit ad hoc. Refactor it somewhere?


codyroux:
Looking good other than the conflicts! I agree that the smtlib stuff should be in its own module maybe something like `Z3_utils.ml`. I'll take a closer look in the morning.

wp/lib/bap_wp/src/environment.mli
```diff
@@ -192,4 +195,20 @@ val mk_z3_expr : Z3.context -> name:string -> typ:Bap.Std.Type.t -> Constr.z3_ex
     constant is "fresh" with the {!Environment.var_gen}. *)
 val new_z3_expr : ?name:string -> t -> Bap.Std.Type.t -> Constr.z3_expr
 
+
+(** [get_decls_and_symbols] builds from a the var_map in an environment 
+    a mapping of all Z3 func_decl to their symbol. This is a helper function for
+    [mk_smtlib2] *)
+
```
fortunac: Just a little nitpick, but there are no blank lines between the ocamldoc comments and the function signatures in the other functions, and I think it would be great to be consistent.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -217,11 +217,10 @@ let exp_to_z3 (exp : Exp.t) (env : Env.t) : Constr.z3_expr * Constr.t list * Con
...
         (Var.to_string v) (Exp.to_string exp) (Exp.to_string body);
       let exp_val, env = exp_to_z3_body exp env in
-      (* FIXME: we're handling this incorrectly! The variable should
-         be removed from the context after leaving the scope of the
-         Let! *)
       let env' = Env.add_var env v exp_val in
-      exp_to_z3_body body env'
+      let z3_expr, env = exp_to_z3_body body env' in
+      let env = Env.remove_var env v in
+      (z3_expr, env)
```
fortunac: If a variable in the let-binding shares a name with a variable that already existed in the `env`, would removing it also remove the original variable? If so, it might be good to write a comment about that.

wp/lib/bap_wp/src/compare.ml
```diff
@@ -116,19 +116,43 @@ let compare_subs_empty_post
...
+  let ctx = Env.get_context env1 in
+  Constr.mk_smtlib2 ctx smtlib_str declsym
+
 let compare_subs_eq
     ~input:(input : Var.Set.t)
     ~output:(output : Var.Set.t)
     ~original:(original : Sub.t * Env.t)
     ~modified:(modified : Sub.t * Env.t)
+    ~smtlib_post:(smtlib_post : string)
+    ~smtlib_pre:(smtlib_pre : string)
```
fortunac: I agree, hypothesis might be a better name for this.

wp/lib/bap_wp/tests/unit/test_compare.ml
```diff
@@ -123,7 +123,7 @@ let test_sub_pair_1 (test_ctx : test_ctxt) : unit =
   let output_vars = Var.Set.singleton z in
   let compare_prop, _, _ = Comp.compare_subs_eq
       ~input:input_vars ~output:output_vars
-      ~original:(sub1,env1) ~modified:(sub2,env2) in
+      ~original:(sub1,env1) ~modified:(sub2,env2) ~smtlib_post:"" ~smtlib_pre:"" in
```
fortunac: Should `smtlib_post` and `smtlib_pre` be optional arguments defaulting to `""`? This one is really up to you.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -833,6 +796,7 @@ let check ?refute:(refute = true) (solver : Solver.solver) (ctx : Z3.context)
     else
       pre'
   in
+  Printf.printf "Z3 Query:\n%s\n" (Z3.Expr.to_string (Z3.Expr.simplify is_correct None));
```
fortunac: Do you think we can make this an `info` log or something so that it doesn't print out when we run the unit tests?

fortunac:
Looks super good! Just a few nitpicks. I agree, putting the smtlib functions in its own module and adding unit test(s) for the compare module is a great idea. Good work! :tada:

wp/lib/bap_wp/src/precondition.ml
codyroux: Correct, this could conceivable occur if some pervert wanted to name a let-bound variable `RAX`, or more pragmatically if an inner let shadows an outer let. This should cause a run-time error rather than a mysterious change in semantics. A comment might be appropriate though.

philzook58: Hmm. That's an interesting point. We didn't implement this correctly basically. We should store the variable we're covering up. In practice it probably isn't a problem because the let bindings seem to use names like "$1", but should probably still be fixed for sanity

wp/lib/bap_wp/src/compare.ml
codyroux: So `smtlib_hyp` or `smtlib_hype` if we're trying to appeal to the youth.

wp/lib/bap_wp/tests/unit/test_compare.ml
philzook58: Yeah. i debated this one. I just hated putting them as the first arguments, which optional seems to require. But it might be the right call

wp/lib/bap_wp/src/precondition.ml
philzook58: Seems reasonable. Or perhaps a command line flag.
Is there an insane amount of stuff in info? I would be of the opinion we also suppress the Constrain being printed by default and maybe the BIL.

wp/lib/bap_wp/src/precondition.ml
philzook58: This is a really good catch

wp/lib/bap_wp/src/precondition.ml
fortunac: I'm not against having a flag for `verbose` mode, but it's probably out of the scope for this PR. And there isn't too much stuff in info. From grepping, I can see we have 8 info logs.

wp/lib/bap_wp/tests/unit/test_compare.ml
codyroux: I vote for `true` and `false` as defaults, respectively.

wp/lib/bap_wp/src/precondition.ml
codyroux: Yea `info` seems reasonable... IDK.

wp/lib/bap_wp/tests/unit/test_compare.ml
philzook58: Shouldn't the defaults both be `true`?

codyroux: Yes, I think you're right, actually.

wp/lib/bap_wp/src/precondition.ml
philzook58: I think I fixed this

wp/lib/bap_wp/src/z3_utils.mli
```diff
@@ -0,0 +1,47 @@
...
+(*                                                                         *)
+(*  This work is funded in part by ONR/NAWC Contract N6833518C0107.  Its   *)
+(*  content does not necessarily reflect the position or policy of the US  *)
+(*  Government and no official endorsement should be inferred.             *)
+(*                                                                         *)
+(***************************************************************************)
+
+(**
+
+   This module exports types and utilities to process and report results found
```
fortunac: Can we update this comment to match the `z3_utils` module?

philzook58: Fixed. Good catch.

PR #107, opened 2020-01-02
Fortunac/memory models2

fortunac:
This is a pretty big one:

Adds to the environment the valid memory ranges for the stack and heap.
  - the default range for stack is: 0x00007fffffff0000 - 0x00007fffffffffff
  - the default range for the heap is: 0x0000000000000000 - 0x00000000ffffffff
  - currently, you can change the ranges in `Pre.mk_env`, but I have not added the ability to change it from the plugin side

Generates fresh variables for the initial states of memory for the original and modified binaries: `init_mem_orig` and `init_mem_mod`

Add hooks to memory reads that assume that:
  - if the address being read is on the heap, then the initial value at that address in the original binary is equal to initial value at that address plus an offset in the modified binary (`Heap(x) => init_mem_orig[x] == init_mem_mod[x + d]`)
  - if the address being read is on the stack, then the initial values at that address in the original binary and the modified binary are equal (`Stack(x) => init_mem_orig[x] == init_mem_mod[x]`)

Adds to the hypothesis when comparing two subs:
  - the fresh variables are the initial states of memory: `init_mem_orig == mem_orig` and `init_mem_mod == mem_mod`
  - the stack pointer is initially pointing to a valid location on the stack
  - removes the hypothesis that state the initial states of memory are equal between the original and modified binaries when we run the memory hooks

Countermodel can now show the memories for both the original and modified binaries

Gives the hooks the option to be added to the postcondition before or after substitution in `visit_elt`
  - the assumptions and VCs for checking non-null reads have been set to `BeforeExec`
  - the assumptions for equating the memory reads between the binaries have been set to `AfterExec`
  - creates a record of type `hooks` which contains all the assumptions and VCs, in order to avoid returning 4 values in `exp_to_z3` and having to keep track which list contains assumptions or VCs
  - I had to change the API for this, so there may be breaking changes between WP and Conc


wp/lib/bap_wp/src/compare.ml
```diff
@@ -35,13 +38,41 @@ let set_to_eqs (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t) : Constr.t list
     ~f:(fun (eqs, env1, env2) v ->
         let var1, env1 = Env.get_var env1 v in
         let var2, env2 = Env.get_var env2 v in
-        let eq = Bool.mk_eq ctx var1 var2
-                 |> Constr.mk_goal
-                   (Format.sprintf "%s = %s" (Expr.to_string var1) (Expr.to_string var2))
-                 |> Constr.mk_constr
-        in (eq::eqs, env1, env2)
+        (* We do not compare memory when we assume that memory is equal at an offset. *)
```
codyroux: Should we *ever* compare memory by default?

fortunac: This is where I was unsure,  but I think it is reasonable to assume that memory is not going to be equal by default. We still get the same results whether we add the memory comparison to the hypothesis or not.

wp/lib/bap_wp/src/environment.ml
```diff
@@ -205,6 +212,29 @@ let init_loop_unfold (num_unroll : int) : loop_handler =
       unroll
   }
 
+let init_mem_range ctx arch (min, max) : Constr.z3_expr -> Constr.z3_expr =
+  let sort = arch |> Arch.addr_size |> Size.in_bits |> BV.mk_sort ctx in
+  let min = Expr.mk_numeral_int ctx min sort in
+  let max = Expr.mk_numeral_int ctx max sort in
+  fun addr -> Bool.mk_and ctx [BV.mk_ule ctx min addr; BV.mk_ule ctx addr max]
+
+(* Creates a new environment with
```
philzook58: Love the documentation!

wp/lib/bap_wp/src/environment.ml
```diff
@@ -64,7 +69,9 @@ and loop_handler = {
   handle : t -> Constr.t -> start:Graphs.Ir.Node.t -> Graphs.Ir.t -> t
 }
 
-and cond_type = Verify of Constr.goal | Assume of Constr.goal
+and cond = BeforeExec of Constr.goal | AfterExec of Constr.goal
+
+and cond_type = Verify of cond | Assume of cond
```
codyroux: I'm starting to worry that these are going to start nesting more and more as we come up with more attributes to apply to our conditions. No action needed now, but something to keep in mind before it gets out of hand.

fortunac: Sounds like a good idea

wp/lib/bap_wp/src/compare.ml
```diff
@@ -35,13 +38,40 @@ let set_to_eqs (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t) : Constr.t list
     ~f:(fun (eqs, env1, env2) v ->
         let var1, env1 = Env.get_var env1 v in
         let var2, env2 = Env.get_var env2 v in
-        let eq = Bool.mk_eq ctx var1 var2
-                 |> Constr.mk_goal
-                   (Format.sprintf "%s = %s" (Expr.to_string var1) (Expr.to_string var2))
-                 |> Constr.mk_constr
-        in (eq::eqs, env1, env2)
+        if Target.CPU.is_mem v then
```
codyroux: :+1:

wp/lib/bap_wp/src/compare.ml
```diff
@@ -35,13 +38,40 @@ let set_to_eqs (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t) : Constr.t list
...
+          eqs, env1, env2
+        else
+          let eq = Bool.mk_eq ctx var1 var2
+                   |> Constr.mk_goal
+                     (Format.sprintf "%s = %s" (Expr.to_string var1) (Expr.to_string var2))
+                   |> Constr.mk_constr
+          in (eq::eqs, env1, env2)
       )
 
+(* Adds the hypothesis: [var == init_var]. *)
```
codyroux: Do we have docs on how to use `foo_init` in specs? Do we have examples? Do we have C examples?

wp/lib/bap_wp/src/compare.ml
```diff
@@ -35,13 +38,40 @@ let set_to_eqs (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t) : Constr.t list
...
+      init env1 v "orig" :: init env2 v "mod" :: init_vars)
+
+(* Adds hypothesis that the stack pointer is within the valid range of the stack
+   according to the environment. *)
+let set_sp_range (env : Env.t) : Constr.t =
+  let arch = Env.get_arch env in
+  let module Target = (val target_of_arch arch) in
+  let sp, _ = Env.get_var env Target.CPU.sp in
+  Env.in_stack env sp
+  |> Constr.mk_goal "SP points to stack"
```
codyroux: Maybe, `"SP within stack"`, or `"Initial SP constraint"`?

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -49,14 +49,21 @@ let get_model_exn (solver : Solver.solver) : Model.model =
...
 
-let format_values (fmt : Format.formatter) (vals : (Var.t * z3_expr) list) : unit =
-  List.iter vals
-    ~f:(fun (var, value) ->
-        let var_str = Var.to_string var in
-        let pad_size = Int.max (5 - (String.length var_str)) 1 in
-        let pad = String.make pad_size ' ' in
-        Format.fprintf fmt
-          "\t%s%s|->  @[%s@]@\n" var_str pad (Expr.to_string value))
+let format_values (fmt : Format.formatter) (vals : (string * z3_expr) list) : unit =
```
codyroux: :+1:

wp/lib/bap_wp/src/environment.ml
```diff
@@ -42,7 +44,10 @@ type t = {
   loop_handler : loop_handler;
   exp_conds : exp_cond list;
   arch : Arch.t;
-  fun_input_regs : bool
+  fun_input_regs : bool;
+  stack : Constr.z3_expr -> Constr.z3_expr;
```
codyroux: What's the input to this feller? Comments here might be useful.

wp/lib/bap_wp/src/environment.ml
```diff
@@ -42,7 +44,10 @@ type t = {
   loop_handler : loop_handler;
   exp_conds : exp_cond list;
   arch : Arch.t;
-  fun_input_regs : bool
+  fun_input_regs : bool;
```
codyroux: Mabye a more descriptive name here, e.g. starting with a verb, like "are" or "has".

wp/lib/bap_wp/src/compare.ml
fortunac: Nope, but I can definitely write up some documentation on the init vars (in the .mli?) and a little test example.

wp/lib/bap_wp/src/compare.ml
fortunac: Yay, better names!

wp/lib/bap_wp/src/environment.ml
fortunac: Can do.

wp/lib/bap_wp/src/environment.ml
fortunac: :+1:

wp/lib/bap_wp/src/environment.mli
```diff
@@ -83,6 +89,9 @@ type exp_cond = t -> Bap.Std.Exp.t -> cond_type option
     - the architecture of the binary
     - the option to freshen variable names
     - the option to use all input registers when generating function symbols at a call site
+    - the range of addresses of the stack
```
codyroux: The *concrete* range of addresses. We might want to make these more abstract at some point, but let's keep it for now.

wp/lib/bap_wp/src/environment.mli
```diff
@@ -138,7 +149,7 @@ val add_precond : t -> Bap.Std.Tid.t -> Constr.t -> t
 
 (** Creates a verifier checker for a {!Constr.z3_expr}, returning first the assumptions, then the
```
codyroux: Maybe fix this comment.

wp/lib/bap_wp/src/environment.mli
```diff
@@ -201,6 +212,27 @@ val is_x86 : Bap.Std.Arch.t -> bool
     when generating symbols in the function specs at a function call site. *)
 val use_input_regs : t -> bool
 
+(** Obtains a Z3 expression that checks if an address is within the stack region
+    of memory: i.e. `stack_min <= addr <= stack_max`. *)
```
codyroux: Ok here's the doc. Maybe be more explicit, e.g. "`in_stack env addr` is the constraint `STACK_MIN <= addr <= STACK_MAX` as defined by the `env`."

wp/lib/bap_wp/src/environment.mli
```diff
@@ -201,6 +212,27 @@ val is_x86 : Bap.Std.Arch.t -> bool
...
+(** [mk_init_var env var suffix] creates a fresh Z3 variable that represents the
+    initial state of variable [var] with suffix 'orig' for the original binary
+    and 'mod' for the modified binary. *)
+val mk_init_var : t -> Bap.Std.Var.t -> string -> Constr.z3_expr
+
+(** Add a new binding to the environment for a bap variable to a Z3 expression
+    that represents its initial state. *)
+val set_init_var : t -> Bap.Std.Var.t -> Constr.z3_expr -> t
+
+(** [get_init_var var] obtains the Z3 expression that represents the initial state
```
codyroux: :+1:

wp/lib/bap_wp/src/environment.mli
```diff
@@ -201,6 +212,27 @@ val is_x86 : Bap.Std.Arch.t -> bool
...
+val in_heap : t -> Constr.z3_expr -> Constr.z3_expr
+
+(** [mk_init_var env var suffix] creates a fresh Z3 variable that represents the
+    initial state of variable [var] with suffix 'orig' for the original binary
+    and 'mod' for the modified binary. *)
+val mk_init_var : t -> Bap.Std.Var.t -> string -> Constr.z3_expr
+
+(** Add a new binding to the environment for a bap variable to a Z3 expression
+    that represents its initial state. *)
+val set_init_var : t -> Bap.Std.Var.t -> Constr.z3_expr -> t
```
codyroux: Do we need both `mk_init_var` and `set_init_var`? Do we ever call one without the other?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -594,14 +632,15 @@ let visit_jmp (env : Env.t) (post : Constr.t) (jmp : Jmp.t) : Constr.t * Env.t =
...
-              (Expr.to_string cond_val) (List.to_string ~f:Constr.to_string vcs)
-              (List.to_string ~f:Constr.to_string assume);
+            let cond_val, hooks, env = exp_to_z3 cond env in
+            debug "\n\nJump when %s:\n%s\n%!"
+              (Expr.to_string cond_val) (hooks_to_string hooks);
             let cond_size = BV.get_size (Expr.get_sort cond_val) in
             let false_cond = Bool.mk_eq ctx cond_val (z3_expr_zero ctx cond_size) in
             let ite = Constr.mk_ite jmp (Bool.mk_not ctx false_cond) l_pre post in
-            let post = ite::vcs in
+            let vcs = hooks.verify_before @ hooks.verify_after in
```
codyroux: Maybe we should be a little careful here: if we add a `PC` variable, these won't be identical anymore...

codyroux: Not sure what the fix is, other than adding a comment...

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -651,16 +690,21 @@ let visit_elt (env : Env.t) (post : Constr.t) (elt : Blk.elt) : Constr.t * Env.t
...
-    Constr.substitute_one post z3_var rhs_exp, Env.add_var env var z3_var
+    (* Adding the specified assumptions and VCs to the postcondition before applying
+       the substitution. *)
+    let post = post :: hooks.verify_before in
+    let post = Constr.mk_clause hooks.assume_before post in
+    let post = Constr.substitute_one post z3_var rhs_exp in
+    (* Adding the specified assumptions and VCs to the postcondition after applying
+       the substitution. *)
+    let post = post :: hooks.verify_after in
+    let post = Constr.mk_clause hooks.assume_after post in
```
codyroux: :+1:

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -840,16 +885,67 @@ let non_null_vc : Env.exp_cond = fun env exp ->
   if List.is_empty conds then
     None
   else
-    Some (Verify (Constr.mk_goal "verify" (Bool.mk_and ctx conds)))
+    Some (Verify (BeforeExec (Constr.mk_goal "verify" (Bool.mk_and ctx conds))))
```
codyroux: :+1:

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -840,16 +885,67 @@ let non_null_vc : Env.exp_cond = fun env exp ->
...
 let non_null_assert : Env.exp_cond = fun env exp ->
   let ctx = Env.get_context env in
   let conds = collect_non_null_expr env exp in
   if List.is_empty conds then
     None
   else
-    Some (Assume (Constr.mk_goal "assume" (Bool.mk_and ctx conds)))
+    Some (Assume (BeforeExec (Constr.mk_goal "assume" (Bool.mk_and ctx conds))))
+
+(* At a memory read, add two VCs of the form:
```
codyroux: Surely, these are assumptions, not VCs right?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -840,16 +885,67 @@ let non_null_vc : Env.exp_cond = fun env exp ->
...
+  let module Target = (val target_of_arch arch) in
+  let init_mem1 = Env.mk_init_var env1 Target.CPU.mem "orig" in
+  let init_mem2 = Env.mk_init_var env2 Target.CPU.mem "mod" in
+  let env1 = Env.set_init_var env1 Target.CPU.mem init_mem1 in
+  let env2 = Env.set_init_var env2 Target.CPU.mem init_mem2 in
+  let conds = collect_mem_read_expr env1 env2 exp offset in
+  if List.is_empty conds then
+    None
+  else
+    Some (Assume (AfterExec (Constr.mk_goal "assume" (Bool.mk_and ctx conds))))
```
codyroux: Is there no better name for this goal?

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -60,7 +65,8 @@ val get_vars : Bap.Std.Sub.t -> Bap.Std.Var.Set.t
     given the name of each variable. *)
 val get_output_vars : Bap.Std.Sub.t -> string list -> Bap.Std.Var.Set.t
 
-(** Create a Z3 expression that denotes a load in memory [mem] at address [addr]. *)
+(** Create a Z3 expression that denotes a load in memory [mem] at address [addr]
```
codyroux: Can we describe each input in the comments, even if the are somewhat self evident?

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -34,6 +34,11 @@ module Env = Environment
 
 module Constr = Constraint
 
+(** A record of generated constraint lists mapped to assumptions and verification
```
codyroux: If this type is opaque, arguably we need to be *less* descriptive about the insides, and talk more about what a user needs to do with it...

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -177,6 +188,9 @@ val num_unroll : int ref
     - an architecture of x86_64
     - freshening variables set to false
     - and using all function input registers when generating function symbols at a call site
+    - the default range of addresses of the stack: [0x00007fffffff0000, 0x00007fffffffffff]
```
codyroux: Maybe we should describe also what these inputs are supposed to do rather than just their default values. It seems tedious though so I won't insist too much.

wp/plugin/wp.ml
```diff
@@ -97,6 +95,17 @@ let analyze_proj (proj : project) (var_gen : Env.var_gen) (ctx : Z3.context)
     (Sub.to_string main_sub) Constr.pp_constr pre;
   (pre, env, env)
 
+(* If an offset is specified, generates a function of the address of a memory read in
+   the original binary to the address plus an offset in the modified binary. *)
+let get_exp_conds (env : Env.t) (offset : int) : Env.exp_cond list =
+  let ctx = Env.get_context env in
+  let addr_offset = fun addr ->
+    let width = addr |> Z3.Expr.get_sort |> Z3.BitVector.get_size in
+    let offset = Z3.BitVector.mk_numeral ctx (string_of_int offset) width in
```
codyroux: Isn't there a better function for this?

codyroux: Nevermind, there isn't. Sheesh.

codyroux:
Modulo some small comment related details, looks good to me. Ready to go when you are.

Issue #113, opened 2020-01-21
Constraint evaluation slow or looping on big examples


philzook58:
I'm running the regular wp on main from base64-gcc from the retrowrite examples using comparison on literally identical project files. 

It appears to stuck in `Constr.eval` inside `Pre.check`. It is processing things, but there are just a ton of constraints it is going through. If I set `--wp-output-vars=[]` then I can get it to finish. I was seeing many goals created of the form "RAX0 = RAX0167" and not many others. To be expected?

philzook58:
Chloe seems to have vastly improved the constraint evaluation time with some relatively small changes.
Perhaps close this?

codyroux:
Let's wait till she returns maybe...

PR #114, opened 2020-02-05
Remove implication for empty hypotheses when evaluating clauses

fortunac:
Addresses #113. When comparing `main` with itself in our retrowrite example, evaluation time from `Constr.t` to `z3_expr` dropped from ~600 seconds to ~30 seconds.


codyroux:
If only all merge commits were this easy to review...

PR #118, opened 2020-02-20
Fortunac/retrowrite mem

fortunac:
Added a small module `Symbols` which utilizes objdump to find the starting address for symbols in the data and bss sections of elf binaries. Uses this to align the memory between the original and modified binary. 

Eventually, I think the ideal way to handle this would be to use the elfsymbols util in conjunction with BAP's knowledge base, but I think this module is a good start.


philzook58:
Looks good. Tests pass and makes sense. I wonder if it would be possible to decouple the z3 bits from the objdump bits more? It strikes me as odd that z3 is involved at all in functions that are calling and parsing objdump results.

I think there could be functions that roughly looks more like

```
type addr = int (* maybe a string of the adress? *)
(* returns dictionary associating address to symbol *)
val get_symbols : string -> (symbol * addr) list  (* or use a Map from symbol to addr *)
val offset_constraint : ~orig:(symbol * addr) list -> ~mod : (symbol * addr) list -> Z3.ctx -> Z3.constr
```

fortunac:
Sounds like a good idea to separate the Z3 bits and the objdump bits. How about we represent the address as the a `Bap.Std.Word.t`?

I chose to use a list instead of a map because I wanted the match symbol names on their prefixes rather than the entire key itself because names have been modified in retrowrite, ex: `foo` vs `foo_1234`. If there is a way to match on prefixes in a map, I'm all for it.

fortunac:
A good example of this use case is in the subroutine `isbase64` in retrowrite which you can run with:
```
bap base64-gcc --pass=wp \
  --wp-compare \
  --wp-file1=base64-gcc.bpj \
  --wp-file2=base64-retrowrite.bpj \
  --wp-output-vars=RAX,RBX,RCX,RDX,RSI,RDI,RSP,RBP,R8,R9,R10,R11,R12,R13,R14,R15 \
  --wp-function=isbase64 \
  --wp-mem-offset=true
```
with an expected output of `UNSAT`.

If you set `--wp-mem-offset=false`, you should get `SAT`.

codyroux:
I agree with Phil's comments, and Chloe's responses, as usual. There is no easy way to match on a prefix of a key to my knowledge, though if one wanted to the appropriate datatype would be a *trie*: https://www.lri.fr/~filliatr/ftp/ocaml/ds/trie.mli.html. I'd advise strongly against it for now.

fortunac:
We can potentially use https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/Map/Tree/index.html#val-subrange as a way to search up prefixes, but let's hold off on it for now.

PR #122, opened 2020-03-20
bumped core version to 0.12.4 in opam files

DieracDelta:
The opam files also need to bump the core dependency 0.11 to 0.12.4. This PR performs this bump.


fortunac:
Thanks!

PR #125, opened 2020-03-20
Fortunac/mem read fix

fortunac:
1.  At a memory read, rather than generating a constraint of 
```
(Heap(x) => init_mem_orig[x] = init_mem_mod[x + d]) ^ (Stack(x) => init_mem_orig[x] = init_mem_mod[x])
```
we generate a constraint in the form of
```
if stack(x) then
  init_mem_orig[x] = init_mem_mod[x]
else if heap(x) then
  init_mem_orig[x] = init_mem_mod[x + d]
else
  init_mem_orig[x] = init_mem_mod[x]
```
in order to cover all of memory.

2. Fix bug where `__VERIFIER_nondet` was not freshening the variable we are chaosing.

3. Add a stub spec for `calloc`, which has the same behavior as `__VERIFIER_nondet`.

### DIFF
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -480,6 +480,31 @@ let spec_verifier_nondet (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
   else
     None
 
+let spec_calloc (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
+  if String.equal (Sub.name sub) "calloc" then
+    let open Env in
+    Some {
+      spec_name = "spec_calloc";
+      spec = Summary
+        (* Currently stubbing this spec to mirror that of spec_verifier_nondet. *)
+        (fun env post tid ->
+             let post = set_fun_called post env tid in
+             let post, env = increment_stack_ptr post env in
+             let args = Term.enum arg_t sub in
+             let output =
+               match Seq.find args ~f:(fun a -> Bap.Std.Arg.intent a = Some Bap.Std.Out) with
+               | Some o -> o
+               | None -> failwith "Verifier headerfile must be specified with --api-path" in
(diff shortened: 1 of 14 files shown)


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -472,7 +472,35 @@ let spec_verifier_nondet (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
...
+  else
+    None
+
+let spec_calloc (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
+  if String.equal (Sub.name sub) "calloc" then
+    let open Env in
+    Some {
+      spec_name = "spec_calloc";
+      spec = Summary
+        (* Currently stubbing this spec to mirror that of spec_verifier_nondet. *)
```
codyroux: Are we getting code duplication from `spec_verifier_nondet`?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -472,7 +472,35 @@ let spec_verifier_nondet (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
...
-             subst_fun_outputs env sub post ~inputs:[] ~outputs:[v], env)
+             let z3_v, env = Env.get_var env v in
+             let name = Format.sprintf "%s_ret_%s" (Sub.name sub) (Expr.to_string z3_v) in
+             let fresh = new_z3_expr env ~name:name (Var.typ v) in
+             Constr.substitute_one post z3_v fresh, env)
+    }
+  else
+    None
+
+let spec_calloc (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
```
codyroux: Should we also have `malloc` while we're at it?

wp/lib/bap_wp/src/precondition.ml
fortunac: That is true. I can add the calloc and malloc case to `nondet`'s spec.

Issue #123, opened 2020-03-20
Plugin Tests


fortunac:
Update plugin tests:
- We currently have `run_wp.sh` that aren't being tested
- The default values built in `test_wp.ml` do not match the default flag options in the plugin itself.

The solution would be to stop building the command line options, and to invoke the `run_wp.sh` scripts directly instead.

We should also make sure all of the sample binaries are being run in our tests.

fortunac:
Closed with #128.

PR #128, opened 2020-03-25
Fortunac/plugin tests

fortunac:
Refers to #123.

Big change to `test_wp.ml`. Instead of building a command in the test file, we run the scripts from the `sample_binaries` directly. That way we can actually test our plugin's default flags and test that our scripts work. There used to be a mismatch of default flags when we were building the commands in the test file.


wp/resources/sample_binaries/retrowrite_stub/run_wp_inline_afl.sh
```diff
@@ -0,0 +1,26 @@
+# Stubs the lines of asssembly that retrowrite addsto the beginning of each
```
codyroux: did you run spell-check on these?

wp/resources/sample_binaries/retrowrite_stub_no_ret/run_wp.sh
```diff
@@ -1,3 +1,12 @@
+# Stubs the lines of asssembly that retrowrite adds to the beginning of each
```
codyroux: Typo

codyroux:
I'm very impressed, this is good work and should result in a large quality of life improvement.

wp/resources/sample_binaries/retrowrite_stub/run_wp_inline_afl.sh
fortunac: Apparently not. :)

PR #133, opened 2020-04-08
Use regex to match symbols

fortunac:
Fixes the finicky prefix hack we used to match symbols from the original binary to the modified binary.


codyroux:
Looks reasonable, I'm sure this will fail sometime in the future, but we'll let future @fortunac and @codyroux worry about it.

Issue #132, opened 2020-04-06
--wp-inline flag not catching inlined function calls between `.so` files 


DieracDelta:
I've written two `.so` files and I'm trying to use CBAT to catch the differences in `example`. `main_1` looks like:
```
#include <stdint.h>
#include <stdlib.h>
int8_t init(){
        return 0;
}

int8_t example(){
        int8_t err = init();
        return err;
}
```

`main_2` looks like:
```
#include <stdint.h>
#include <stdlib.h>

int8_t init(){
        return 1;
}

int8_t example(){
        int8_t err = init();
        return err;
}
```
The expected result when diffing these `example` in these two libraries is SAT  because `init()` returns a different value in each library. I compile and run with the following script:
```
run() {
        gcc -fPIC -fno-stack-protector -shared -Wall -o main_1.so main_1.c
        gcc -fPIC -fno-stack-protector -shared -Wall -o main_2.so main_2.c
        bap main_1.so --pass=save-project --save-project-filename=main_1.bpj
        bap main_2.so --pass=save-project --save-project-filename=main_2.bpj
        bap /bin/echo --pass=wp --wp-compare=true --wp-file1=main_1.bpj --wp-file2=main_2.bpj --wp-function=example --wp-inline="init"
}

clean() {
        rm main_1.so main_2.so main_1.bpj main_2.bpj
}

clean ; run
```
This returns UNSAT. The dot files for `example` in `main_1` and `main_2` are identical and the following:
```
000005b7: program
000005a0: sub example()
000003b8: 
000003bf: #44 := RBP
000003c2: RSP := RSP - 8
000003c5: mem := mem with [RSP, el]:u64 <- #44
000003cc: RBP := RSP
000003da: #45 := RSP
000003dd: RSP := RSP - 0x10
000003e0: CF := #45 < 0x10
000003e3: OF := high:1[(#45 ^ 0x10) & (#45 ^ RSP)]
000003e6: AF := 0x10 = (0x10 & (RSP ^ #45 ^ 0x10))
000003e9: PF := ~low:1[let $1 = RSP >> 4 ^ RSP in let $2 = $1 >> 2 ^ $1 in
$2 >> 1 ^ $2]
000003ec: SF := high:1[RSP]
000003ef: ZF := 0 = RSP
000003f6: RAX := 0
000003ff: RSP := RSP - 8
00000402: mem := mem with [RSP, el]:u64 <- 0x6AD
00000405: call @init with return %00000407

[...]

PR #134, opened 2020-04-10
Add spec for inlining plt entries

fortunac:
Refers to the bug found in #132

### DIFF
--- a/wp/lib/bap_wp/src/environment.ml
+++ b/wp/lib/bap_wp/src/environment.ml
@@ -52,7 +52,7 @@ type t = {
 
 and fun_spec_type =
   | Summary of (t -> Constr.t -> Tid.t -> Constr.t * t)
-  | Inline
+  | Inline of Tid.t
 
 and fun_spec = {
   spec_name : string;
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -91,7 +91,7 @@ let lookup_sub (label : Label.t) (post : Constr.t) (env : Env.t) : Constr.t * En
     begin
       match Env.get_sub_handler env tid with
       | Some (Summary compute_func) -> compute_func env post tid
-      | Some Inline -> !inline_func post env tid
+      | Some (Inline target) -> !inline_func post env target
       | None -> post, env
     end
   (* TODO: Evaluate the expression for the indirect jump and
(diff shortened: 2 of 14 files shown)


DieracDelta:
I tried this on a couple of examples and it works great!

wp/lib/bap_wp/src/environment.ml
```diff
@@ -52,7 +52,7 @@ type t = {
 
 and fun_spec_type =
   | Summary of (t -> Constr.t -> Tid.t -> Constr.t * t)
-  | Inline
+  | Inline of Tid.t
```
codyroux, line 55: This is actually quite nice!

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -626,11 +626,29 @@ let spec_inline (to_inline : Sub.t Seq.t) (sub : Sub.t) (_ : Arch. t)
   if Seq.mem to_inline sub ~equal:Sub.equal then
     Some {
       spec_name = "spec_inline";
-      spec = Inline
+      spec = Inline (Term.tid sub)
     }
   else
     None
 
+let spec_inline_plt (to_inline : Sub.t Seq.t) (sub : Sub.t) (_ : Arch.t)
```
codyroux, line 634: Comments here might be quite welcome, when we debug this in 3 months...

codyroux:
Looks good! Modulo some explanation of our strategy for finding the body of `plt` function calls.

wp/lib/bap_wp/src/precondition.ml
ivg: This is not a PLT pattern, neither it is guaranteed that a function that has `@` symbol is a PLT (or any other) stub, nor that the `@` symbol will be used in the future. So this code is quite brittle and most likely will break in the future. 

The long story. In order to maintain the injectivity of the tied to name mapping, we have to ensure that all functions also have unique names. This invariant is maintained on the `Bap.Std.Program` level, and when a new function is added we check that it is unique and if it is not we mangle its name (using `@<addr>` is just one of the strategies that is used). See https://github.com/BinaryAnalysisPlatform/bap/issues/808 for more information on that, and here is the [code][1] that performs mangling. As you can see, the strategy can't guarantee that in case of two functions having the same name, we give the mangled name to the stub.

You might also interested in the [stub-resolver][2] ABI pass that @gitoleg is implementing as a part of  https://github.com/BinaryAnalysisPlatform/bap/pull/1036/ (Oleg, btw can you push this pass as a separate PR? As there are some questions about x86/primus part of that PR, but this particular module is independent of that and can be already merged.)



[1]: https://github.com/BinaryAnalysisPlatform/bap/blob/master/lib/bap_types/bap_ir.ml#L308-L341
[2]: https://github.com/BinaryAnalysisPlatform/bap/blob/84e74db3ae6fbf58c878d91eff7745a7700ef877/plugins/stub_resolver/stub_resolver_main.ml

ivg:
Ok, we have cleaned up https://github.com/BinaryAnalysisPlatform/bap/pull/1036, if you guys can pull it and verify that it rectifies issue #132 it would be wonderful.

fortunac:
> Ok, we have cleaned up [BinaryAnalysisPlatform/bap#1036](https://github.com/BinaryAnalysisPlatform/bap/pull/1036), if you guys can pull it and verify that it rectifies issue #132 it would be wonderful.

Thanks a bunch! Will try it now.

codyroux:
Let's close this for now, since it's likely the fix will be tied to some things on the BAP side which are not yet pushed.

ivg:
yep, your tests already passing on our side. I think we will push it soonish

PR #139, opened 2020-04-17
Fortunac/model printing

fortunac:
In the function spec for nondeterministic functions, we generate a fresh constant variable that represents the output of the function call (i.e. `__VERIFIER_nondet_int_ret_RAX039`). We then save this variable in a set in the environment, and iterate over the set when printing out the model.

For example, the output for `equiv_null_check` is:

```
Model:
	ZF  |->  #b0
	SF  |->  #b0
	RSP |->  #x00007fffffff0c00
	RSI |->  #x0000000000000000
	RDX |->  #x0000000000000000
	RDI |->  #x0000000000000000
	RCX |->  #x0000000000000000
	RBP |->  #x0000000000000000
	RAX |->  #x0000000000000000
	R9  |->  #x0000000000000000
	R8  |->  #x0000000000000000
	PF  |->  #b0
	OF  |->  #b0
	CF  |->  #b0
	AF  |->  #b0
	mem_orig |-> [
		else |-> #x00]
	mem_mod = mem_orig
	malloc_ret_RAX053   |->  #x00007fdfff7efff3
	malloc_ret_RAX02252 |->  #x0000000000000000
```

### DIFF
--- a/wp/lib/bap_wp/src/compare.ml
+++ b/wp/lib/bap_wp/src/compare.ml
@@ -97,8 +97,8 @@ let compare_subs
   let hyps, env1, env2 =
     hyps ~original:(sub1, env1) ~modified:(sub2, env2) ~rename_set:vars in
   info "\nHypotheses:\n%s\n%!" (Constr.to_string hyps);
-  let pre_mod, _ = Pre.visit_sub env2 post sub2 in
-  let pre_combined, _ = Pre.visit_sub env1 pre_mod sub1 in
+  let pre_mod, env2 = Pre.visit_sub env2 post sub2 in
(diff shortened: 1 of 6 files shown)


codyroux:
Seems fine, except for the rather large caveat that I don't understand anything happening with `env`s anymore. When do we pass them through? When do we share them? I have no idea.

But this is an issue for another PR.

fortunac:
We were initially throwing away the updates to each `env` after visiting each subroutine, so we were actually only passing in the envs that were updated from creating the postconditions and hypotheses.

In the single program case, we keep the updated env after visiting a sub and pass that to the Z3 checker. In the comparative case, we we keep the updates to `env1` and `env2` separately and pass that back to the plugin.

PR #143, opened 2020-04-27
Move hypotheses to generic sub comparator

fortunac:
- The default postcondition of wp compares the output registers.
- If, `--wp-check-calls` is set, we don't compare the output registers. Instead we have the postcondition: If a function was not called in the original, then it should not be called in the modified.

- You can add hooks in both cases
- Moved two hypotheses to the generic sub comparator that will be added no matter what case we are in:
  - SP points to a location on the stack
  - all variables are equal to their initial value at the beginning of a program (i.e. RDI = init_RDI)
- You can now pass in preconditions and postconditions via the command line in both cases

### DIFF
--- a/wp/lib/bap_wp/src/compare.ml
+++ b/wp/lib/bap_wp/src/compare.ml
@@ -56,80 +56,6 @@ let init_vars (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t)
   let inits2, env2 = Pre.init_vars vars env2 in
   (inits1 @ inits2), env1, env2
 
-let compare_blocks
-    ~input:(input : Var.Set.t)
-    ~output:(output : Var.Set.t)
-    ~original:(blk1, env1 : Blk.t * Env.t)
-    ~modified:(blk2, env2 : Blk.t * Env.t)
-  : Constr.t * Env.t * Env.t =
-  (* We only freshen variables in blk2, leaving those of blk1 with
-     their original names. *)
-  let env2 = Env.set_freshen env2 true in
-  let output_eq_list, env1, env2 = set_to_eqs env1 env2 output in
-  let output_eq = Constr.mk_clause [] output_eq_list in
-  let input_eq_list, env1, env2 = set_to_eqs env1 env2 input in
-  let pre1, _ = Pre.visit_block env1 output_eq blk1 in
-  let pre2, _ = Pre.visit_block env2 pre1 blk2 in
-  let goal = Constr.mk_clause input_eq_list [pre2] in
-  goal, env1, env2
-
-(* The type of functions that generate a postcondition or hypothesis for comparative
-   analysis. Also updates the environments as needed. *)
-type comparator = original:(Sub.t * Env.t) -> modified:(Sub.t * Env.t) ->
-  rename_set:Var.Set.t -> Constr.t * Env.t * Env.t
-
-(* A generic function for generating a predicate that compares two subroutines.
-   Takes as input a postcondition generator, and a hypothesis generator.
-   this could be made even more generic, say by abstracting over the type of terms
-   being compared. *)
-let compare_subs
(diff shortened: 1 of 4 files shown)


wp/lib/bap_wp/src/compare.ml
```diff
@@ -56,21 +56,45 @@ let init_vars (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t)
...
+  let declsym = declsym1 @ declsym2 in
+  let ctx = Env.get_context env1 in
+  Z3_utils.mk_smtlib2 ctx smtlib_str declsym
+
 let compare_blocks
     ~input:(input : Var.Set.t)
     ~output:(output : Var.Set.t)
     ~original:(blk1, env1 : Blk.t * Env.t)
     ~modified:(blk2, env2 : Blk.t * Env.t)
+    ~smtlib_post:(smtlib_post : string)
```
codyroux, line 84: So what happens if the strings are empty?

wp/lib/bap_wp/src/compare.ml
```diff
@@ -56,21 +56,45 @@ let init_vars (env1 : Env.t) (env2 : Env.t) (vars : Var.Set.t)
...
   let output_eq_list, env1, env2 = set_to_eqs env1 env2 output in
-  let output_eq = Constr.mk_clause [] output_eq_list in
+  let smtlib_post = mk_smtlib2_compare env1 env2 smtlib_post in
+  let output_eq = Constr.mk_clause [] (smtlib_post :: output_eq_list) in
   let input_eq_list, env1, env2 = set_to_eqs env1 env2 input in
+  let smtlib_hyp = mk_smtlib2_compare env1 env2 smtlib_hyp in
   let pre1, _ = Pre.visit_block env1 output_eq blk1 in
   let pre2, _ = Pre.visit_block env2 pre1 blk2 in
-  let goal = Constr.mk_clause input_eq_list [pre2] in
+  let goal = Constr.mk_clause (smtlib_hyp :: input_eq_list) [pre2] in
```
codyroux, line 97: We probably don't want to add anything if the `smtlib_hyp` is trivial, since this may be a performance issue, if I remember correctly. Probably not that big a deal though.

wp/lib/bap_wp/src/compare.ml
fortunac: A good question. I will test this right now.

codyroux:
General word of warning: if we ever get this BAP Lisp stuff up and running, all of these `string` smt-lib pre and post conditions need to be replaced by more general constraints.

Not sure if we can do some engineering to make this easier.

fortunac:
> General word of warning: if we ever get this BAP Lisp stuff up and running, all of these `string` smt-lib pre and post conditions need to be replaced by more general constraints.
> 
> Not sure if we can do some engineering to make this easier.

This is true. Was the plan to use BAP Lisp to write all of the properties? Would we still need all of the separate comparison functions in the future if that's the case?

codyroux:
I think we'll want to gradually replace them, but the first step will be replacing just the smt-lib ones.

wp/lib/bap_wp/src/compare.ml
fortunac: Empty strings get translated to empty clauses `() => ()` which evaluates to true. It seems that we have a few of them in the constraints we are building, but I'm unsure where all of them are being generated. I'm wondering if there is a big performance impact from this.

codyroux: Do we ever simplify the full constraint before passing it to Z3?

fortunac: Oh my, we are not.

codyroux: Not that big a deal. I'll merge this, and we can make this a separate issue.

PR #145, opened 2020-04-29
Add tests that demonstrate pic and non-pic

fortunac:
With BAP's new stub resolver, we don't need the PLT spec, and can just inline the target function itself. This PR adds two tests that should work with the resolver, one with PIC enabled and another with PIC disabled.

Addresses #132.


codyroux:
I'm actually a little confused: when would this test fail (if we weren't handling pic correctly)?

fortunac:
With BAP's new stub resolver, these tests should pass.

codyroux:
Great. But do they fail if the stub resolver doesn't work?

fortunac:
Ah yeah, they would fail.

codyroux:
Ok I think I get it.

Issue #146, opened 2020-05-11
Single program analyses with custom smtlib conditions require space character


DieracDelta:
In single program analyses, when including a custom pre or post condition, it seems that any references to `mem` must be followed by a space. This error in the case that `mem` or other custom variables used by CBAT are followed by a closing parenthese or newline.

[Here](https://github.com/draperlaboratory/CBAT-internal/tree/jrestivo/wp_examples/documentation/complex_examples/example_13_8bit) is a relevant example. To run, just run `bash run_wp.sh` with wp installed. The interesting bit is the [post condition](https://github.com/draperlaboratory/CBAT-internal/blob/jrestivo/wp_examples/documentation/complex_examples/example_13_8bit/post_cond.smt):

```
(declare-const x (_ BitVec 64))
(assert
 (bvule
  (select  mem 
	   (bvadd (bvsub init_RSP #x000000000000001c) x))
  (select  mem 
	   (bvadd (bvsub init_RSP #x000000000000001b) x))))
```

CBAT runs successfully.  However, I get an error that `mem` is not found when I remove the " " space character following `mem`: 

```
pass "wp" failed in runtime with Z3.Error("(error \"line 7 column 11: unknown constant mem\")\n")
```

Note that this is only a problem for single binary analysis. I believe the function that is causing the error is [here](https://github.com/draperlaboratory/cbat_tools/blob/master/wp/lib/bap_wp/src/z3_utils.ml#L63). The function `~f` in the fold is matching on the variable name used by Z3 (in this case, `mem`) with a space appended on after. If the variable is followed by a closing parenthese or other type of whitespace (for example newline or tab), this snippet of code won't match on the variable. This isn't an issue with comparisons between binaries as a different function is used and looks for a preceding `init_` or an appended `_mod` or `_orig`.

Note that when I remove the space from being concatenated on line 67, I no longer get an error and CBAT works as expected. E.g.

```
~pattern:(Var.name var)
```

However, this fix doesn't account for the case when a user creates a variable name that contains `Var.name` in a custom SMTLIB expression. Instead for single binaries, it might make sense to add in a relatively unique identifier like prepending "VAR" to all custom variables like `mem`, and mention in the user documentation that the user should not create new variables that start with `VAR`. I have no idea how hard this would be to implement. Alternatively, I can add in some documentation about this to the relevant README to warn users to include a space after referencing variables used internally by CBAT.

fortunac:
Handled in #147.

PR #147, opened 2020-05-14
Fortunac/parse smtlib

fortunac:
Should address issue #146 . This tokenizes the smt string using either whitespaces or parentheses as delimiters. And for each token, if it is in the `env`'s `var_map` or `init_var_map`, replaces it with its Z3 name.


wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -60,14 +60,44 @@ let mk_smtlib2 (ctx : Z3.context) (smtlib_str : string) (decl_syms : (Z3.FuncDec
...
+      else
+        token)
+
 let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
   let var_map = Env.get_var_map env in
-  let smt_post = Env.EnvMap.fold var_map ~init:smt_post 
-      ~f:(fun ~key:var ~data:z3_var smt_post ->
-          String.substr_replace_all smt_post ~pattern:((Var.name var) ^ " ") ~with_:((Expr.to_string z3_var) ^ " ")
-        ) in
+  let init_var_map = Env.get_init_var_map env in
```
codyroux, line 85: This definitely won't scale to huge smt-lib inputs (you're iterating over the whole s-expression for each name, instead of iterating once over the expression and looking up the var name). But this shouldn't ever matter.

codyroux:
Looks good. Did we have an issue to extend this to the comparative analysis with `var_orig` and `var_mod` names?

wp/lib/bap_wp/src/z3_utils.ml
fortunac: You are right! Fixing this should be no problem.

fortunac:
We don't have an issue for extending this to `var_orig` and `var_mod` names. I did find a test care for it though. If a user declares a constant that contains `var_orig` in it, it will be overwritten by our parser (i.e. `RDI_orig_1234` gets overwritten to `RDI0_1234`). I can create the issue and work on a fix..

codyroux:
Sounds good. I'll merge this in the meantime.

codyroux:
If @DieracDelta approves.

fortunac:
Sounds good!

Issue #148, opened 2020-05-15
Parsing smtlib strings as pre and postconditions in comparative analyses


fortunac:
In a comparative analysis, when including a custom pre or postcondition, it is possible for a user-declared variable to be overwritten when substituting variable names to include their mangled suffixes. This happens when a variable contains a `Var.name` appended with a `_orig` or `_mod`.

For example, if the user declares `RDI_orig_1234`, it will be overwritten to `RDI0_1234`.

We should extend the fix from #147 to
https://github.com/draperlaboratory/cbat_tools/blob/75cc89f61d8ed1837df2c569e4bf09cc2d44745b/wp/lib/bap_wp/src/compare.ml#L61

fortunac:
Handled in #149.

PR #149, opened 2020-05-15
Tokenize smtlib string in comparative analysis

fortunac:
Should address #148. This PR extends the fix in #147 and tokenizes, then parses through the smtlib string in order to substitute in the z3 names for variables in custom pre and post conditions in a comparative analysis.


codyroux:
Looks good Chloe! I'd feel reassured if there were at least one or two tests, though.

fortunac:
Can do.

Issue #135, opened 2020-04-10
Updates to printing paths


fortunac:
- The printed paths only contain variables that exist in the original binary. For example, if variable `#1` existed in only the modified binary, it would not be printed. This made it difficult to trace through the printed path of the modified binary.

- Although WP first prints out the original path and then the modified path, it is difficult to determine when the switch occurs. Having headers for this would be nice.

- Within a certain path, the order of the jumps printed out seem to be arbitrary. It would be great if it was in order.

fortunac:
Closed with #151.

PR #151, opened 2020-05-26
Fortunac/path printing

fortunac:
Handles #135. 

- Paths in the original binary will only include registers found in the original binary, and paths in the modified binary will only include registers in the modified binary.
- Clearly print out which path belongs to which binary.
- I did remove ITE creation at an unconditional jump in favor or just passing in the precondition of the jump's target, although this was out of scope of the issue. It had a marginal improvement in performance, and helped print the paths in order. If this was a bad idea, I can remove it.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -20,8 +20,6 @@ module BV = Z3.BitVector
 module Model = Z3.Model
 module Solver = Z3.Solver
 
-exception Not_implemented of string
-
 type z3_expr = Expr.expr
 
 type path = bool Jmp.Map.t
@@ -32,14 +30,6 @@ type reg_map = (z3_expr * z3_expr) list Jmp.Map.t
 type goal = { goal_name : string; goal_val : z3_expr }
 
 type refuted_goal = { goal : goal; path : path; reg_map : reg_map }
-
-module ExprSet = Set.Make(
-  struct
-    type t = z3_expr
-    let compare = Expr.compare
-    let sexp_of_t _ = raise (Not_implemented "sexp_of_t for z3_expr not implemented")
-    let t_of_sexp _ = raise (Not_implemented "t_of_sexp for z3_expr not implemented")
-  end)
 
 let eval_model_exn (model : Model.model) (expr : z3_expr) : z3_expr =
   Model.eval model expr true
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -103,26 +103,12 @@ let format_registers (fmt : Format.formatter) (regs : reg_map) (jmp : Jmp.t)
     format_values fmt reg_vals;
     Format.fprintf fmt "\n%!"
 
-(* Divide the path into the original path and the modified path. *)
-let partition_path (p : path) (regs : reg_map) ~orig:(_ : z3_expr Var.Map.t)
-    ~modif:(var_map2 : z3_expr Var.Map.t) : path * path =
-  Jmp.Map.partitioni_tf p
-    ~f:(fun ~key:jmp ~data:_ ->
-        let regs = Jmp.Map.find_exn regs jmp in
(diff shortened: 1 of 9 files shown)


philzook58:
I think this looks good to me.
@codyroux should give a glance at the changes to precondition.ml, but they seem reasonable

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -709,7 +709,13 @@ let visit_jmp (env : Env.t) (post : Constr.t) (jmp : Jmp.t) : Constr.t * Env.t =
               (Expr.to_string cond_val) (hooks_to_string hooks);
             let cond_size = BV.get_size (Expr.get_sort cond_val) in
             let false_cond = Bool.mk_eq ctx cond_val (z3_expr_zero ctx cond_size) in
-            let ite = Constr.mk_ite jmp (Bool.mk_not ctx false_cond) l_pre post in
+            let is_unconditional = Expr.equal cond_val (z3_expr_one ctx cond_size) in
```
codyroux, line 712: I'm not crazy about this: we're turning a BAP expression into a Z3 expression which we're then comparing as a Z3 expression to the bit vector one.

We should try to just stay withing BAP land, it should be more robust and less confusing.

codyroux: `cond` is within scope as well, so it seems equally convenient.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -79,28 +90,60 @@ let format_registers (fmt : Format.formatter) (regs : reg_map) (jmp : Jmp.t)
     format_values fmt reg_vals;
     Format.fprintf fmt "\n%!"
 
-let format_path (fmt : Format.formatter) (p : path) (regs : reg_map)
-    (var_map : z3_expr Var.Map.t) : unit =
+(* Divide the path into the original path and the modified path. *)
+let partition_path (p : path) (regs : reg_map) ~orig:(_ : z3_expr Var.Map.t)
```
codyroux, line 94: This feels a bit hacky, but I assume there's no easy way to split based on `tid`s?

fortunac: I think it might be possible to split based on the `tid`s. One concern I do have is the case that the binaries have the same `tid` for a path. It's unlikely, but it may be possible.

codyroux: I wonder if the Right Thing to Do™ is to carry around the original source along with the constraints/path. But it's ok if we want to defer this work.

fortunac: I was able to pass in the subroutines for both the original and modified binaries to make the arguments more even, but I wonder if just passing in the original source was a better idea.

wp/lib/bap_wp/src/precondition.ml
fortunac: Was able to check if `cond` is unconditional by checking it with `Word.is_one`.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -92,29 +93,45 @@ let format_registers (fmt : Format.formatter) (regs : reg_map) (jmp : Jmp.t)
...
+            match Term.get_attr jmp address with
+            | None ->
+              Format.fprintf fmt "\t%s %s (Address not found)\n%!" jmp_str taken_str
+            | Some addr ->
+              Format.fprintf fmt "\t%s %s (Address: %s)\n%!"
+                jmp_str taken_str (Addr.to_string addr)
+          end;
+          format_registers fmt regs jmp var_map)
+  in
+  let path_orig, path_mod =
```
codyroux, line 122: Maybe a little comment here?

codyroux:
I approve of this PR, as soon as the merge conflicts are resolved.

wp/lib/bap_wp/src/constraint.ml
fortunac: Comment added!

Issue #110, opened 2020-01-13
Update z3 to 4.8.8-1


philzook58:
The opam-repository could be updated so that installation becomes simpler. Also maybe we get more bug fixes.

fortunac:
Updated instructions in READMEs for this in #152.

ivg:
I would suggest using 4.8.8-1, see
* https://github.com/ocaml/opam-repository/pull/16535
* https://github.com/Z3Prover/z3/pull/4468

fortunac:
This is great, Ivan! Thanks for the heads up.

fortunac:
Closing as Z3 is installed by BAP now.

PR #152, opened 2020-05-26
Update package versions in READMEs

fortunac:
Addresses #110. Now that opam has Z3 4.8.8, we can just install Z3 with `opam install z3` similarly to how we install other libraries. This PR reflects this change and updates the version numbers for the libraries we are using.


codyroux:
👍

PR #153, opened 2020-05-29
Run ocamlformat on WP

fortunac:
- `ocamlformat` should be installed with `opam install ocamlformat`
- You can format your code with `make format`

### DIFF
--- a/wp/Makefile
+++ b/wp/Makefile
@@ -2,7 +2,7 @@ BAP_WP = lib/bap_wp
 WP = plugin
 SAMPLE_BINS = resources/sample_binaries
 
-.PHONY: install doc test clean reinstall
+.PHONY: install doc test clean reinstall format
 
 install:
 	$(MAKE) -C $(BAP_WP) $@.local
@@ -10,6 +10,10 @@ install:
 
 doc:
 	$(MAKE) -C $(BAP_WP) $@
+
+format:
+	$(MAKE) -C $(BAP_WP) $@
+	$(MAKE) -C $(WP) $@
 
 test: install
 	$(MAKE) -C $(BAP_WP) $@
--- a/wp/lib/bap_wp/.ocamlformat
+++ b/wp/lib/bap_wp/.ocamlformat
@@ -1,0 +1,2 @@
+doc-comments = before
+doc-comments-val = before
--- a/wp/lib/bap_wp/Makefile
+++ b/wp/lib/bap_wp/Makefile
@@ -25,6 +25,9 @@ test.performance:
 doc:
 	dune build @doc
 
+format:
+	-dune build @fmt --auto-promote
+
 check.installed.findlib:
 	ocamlfind query $(project)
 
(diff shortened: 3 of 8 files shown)


codyroux:
I'm all for this, with the minor caveat that I prefer the docstrings to appear before rather than after a function definition in the `.mli`.

Major caveat: it looks like you have some changes which are *not* a simple application of the code formatter. These should be part of a separate (and much easier to approve) PR.

fortunac:
> I'm all for this, with the minor caveat that I prefer the docstrings to appear before rather than after a function definition in the `.mli`.
> 
> Major caveat: it looks like you have some changes which are _not_ a simple application of the code formatter. These should be part of a separate (and much easier to approve) PR.

I was able to figure out the config that places docstrings before function definitions, and added it to the `.ocamlformat` files. I also removed the result of running the formatter from this PR.

ivg:
I would advise against using ocamlformat as the style that it enforces is pretty destructive to monadic operations, e.g., see how it destroys Core Theory implementation to [an unreadable mess][1] see [core_theory_basic.ml][2] as a particular example. 
```
  let loadw out dir mem key =
    dir                                  (* <--- o_O *)
    >>= fun dir ->
    mem
    >>-> fun ms mem ->
    key
    >>= fun key ->
    let vs = Mem.vals ms in
    let chunk_size = size vs in
    let needed = size out in
    let rec loop chunks loaded =
      if loaded < needed then
        let key = nsucc !!key (loaded / chunk_size) in
        bind (load !!mem key)
        @@ fun chunk -> loop (var chunk :: chunks) (loaded + chunk_size)
      else ite !!dir (concat out (List.rev chunks)) (concat out chunks) in
    loop [] 0
```

See also https://github.com/BinaryAnalysisPlatform/bap/issues/1149

[1]: https://github.com/BinaryAnalysisPlatform/bap/compare/master...ivg:ocamlformatted-compact
[2]: https://github.com/ivg/bap/blob/40412630d50b8cccdb138bf63aee86b2e717f90f/lib/bap_core_theory/bap_core_theory_basic.ml#L55

fortunac:
> I would advise against using ocamlformat as the style that it enforces is pretty destructive to monadic operations, e.g., see how it tunrs the Core Theory implementation to [an unreadable mess](https://github.com/BinaryAnalysisPlatform/bap/compare/master...ivg:ocamlformatted-compact) see [core_theory_basic.ml](https://github.com/ivg/bap/blob/40412630d50b8cccdb138bf63aee86b2e717f90f/lib/bap_core_theory/bap_core_theory_basic.ml#L55) as a particular example.
> 
> ```
>   let loadw out dir mem key =
>     dir                                  (* <--- o_O *)
>     >>= fun dir ->
>     mem
>     >>-> fun ms mem ->
>     key
>     >>= fun key ->
>     let vs = Mem.vals ms in
>     let chunk_size = size vs in
>     let needed = size out in
>     let rec loop chunks loaded =
>       if loaded < needed then
>         let key = nsucc !!key (loaded / chunk_size) in
>         bind (load !!mem key)
>         @@ fun chunk -> loop (var chunk :: chunks) (loaded + chunk_size)
>       else ite !!dir (concat out (List.rev chunks)) (concat out chunks) in
>     loop [] 0
> ```
> 
> See also [BinaryAnalysisPlatform/bap#1149](https://github.com/BinaryAnalysisPlatform/bap/issues/1149)

Thanks for the head up, Ivan! I didn't realize ocamlformat could be destructive to monads. Judging from the linked issue, ocamlformat isn't ready for use. We'll close this PR for now.

Issue #154, opened 2020-05-29
Loop unfolding is not working on most inputs


codyroux:
While loop unfolding works on some very simple tests, anything more complex does not perform correctly.

The relevant bit that implements the unfolding is here:
https://github.com/draperlaboratory/cbat_tools/blob/c48c1dda3b2e4c4265b399dfd70b61853575e68e/wp/lib/bap_wp/src/environment.ml#L195

And the code that invokes it, within a graph traversal is here:
https://github.com/draperlaboratory/cbat_tools/blob/c48c1dda3b2e4c4265b399dfd70b61853575e68e/wp/lib/bap_wp/src/precondition.ml#L803

fortunac:
![image](https://user-images.githubusercontent.com/14669528/83288046-bdde4d00-a1b0-11ea-976b-d2702d033deb.png)
When I run WP on the above subroutine, with a loop unroll count of 1, I expect the postcondition of `x = 3` to be satisfied, but instead, I am getting `x = 5`. I suspect that WP is applying the substitution in the instruction `x := x + 2` multiple times.

codyroux:
Thanks! Can you add this example and the one from the compiled C to the examples?

fortunac:
I wrote this example in BIL in `test_precondition.ml`, so I am unsure how to add it to the examples. But I can make a PR with it added.

codyroux:
Seems like a logical thing to add. And the C code should be easy.

fortunac:
We might want to return the trivial postcondition over here instead of the postcondition of the entire graph:
https://github.com/draperlaboratory/cbat_tools/blob/efb306278b45212f921f4e54d9ea6b0fe1adb82a/wp/lib/bap_wp/src/precondition.ml#L905

nickroess:
Bump on this issue, loop unrolling bug still in effect as discussed yesterday with @codyroux.

Here is another example where loop unrolling fails. In this example, we try to get WP to generate a string that passes  a `strcmp` check. 

```
#include <assert.h>
#include <string.h>


int mystrcmp(char * s1, char * s2){
  int index = 0;
  while (s1[index] != '\0'){
    if (s1[index] != s2[index])
      return 1;
    index++;
  }
  return 0;
}

int mystrcmp2(char *s1, char * s2){
  if (s1[0] != s2[0])
    return 1;
  if (s1[1] != s2[1])
    return 1;
  return 0;
}

char passbuff[8];
int main(int argc, char ** argv){
[...]

PR #155, opened 2020-05-29
Add tests for loop unrolling

fortunac:
Some tests for #154.


codyroux:
Thanks a bunch! I assume you haven't actually added any tests that run, since these both fail?

fortunac:
`test_loop_5` and the `loop` compiled C examples fail. I forgot to add a skip to `test_loop_5`. Will do that now.

codyroux:
All right!

PR #156, opened 2020-06-01
Add test lengths to suite

fortunac:
Adds a timeout of 20 seconds to all plugin tests except for `no_stack_protection` which has a timeout of 60 seconds. The tests for updating the number of loop unrolls has a timeout of 1 second.


philzook58:
Looks good.

Issue #157, opened 2020-06-01
Options to compare function calls and final register values are mutually exclusive


fortunac:
The two options of comparing function calls and comparing final register values are currently mutually exclusive.

https://github.com/draperlaboratory/cbat_tools/blob/87fdb022be3356efc7f44405938a6c901adc1c27/wp/plugin/wp.ml#L188

If a user specifies both `--wp-check-calls` and `--wp-output-vars` from the command line, WP will only compare the function calls. We should add a way in `compare.ml` that allows the user to specify combinations of constraints for the analysis.

Rather than having each function call `compare_subs` as in:

https://github.com/draperlaboratory/cbat_tools/blob/87fdb022be3356efc7f44405938a6c901adc1c27/wp/lib/bap_wp/src/compare.ml#L198

we can have each function generate their postconditions and hypotheses, and then pass it to `compare_subs` outside of the module.

PR #158, opened 2020-06-03
Fortunac/compare flags

fortunac:
Addresses #157. This PR allows the user to specify any combination of:
* `--wp-precond/--wp-postcond`
* `--wp-check-null-derefs`
* `--wp-check-calls`
* `--wp-output-vars`

The way a user invokes each flag remains the same, but the biggest difference is that if a user specifies `check-calls`, wp will still compare the final register values unless the user specifies `--wp-output-vars=""`.

There are major changes to the `compare.ml` module. `compare_subs_fun`, `compare_subs_eq`, `compare_subs_smt`, `compare_subs_empty`, and `compare_subs_sp` each return a pair of `comparator`s which show how to generate the postcondition and hypotheses. From the plugin side, a user can combine these `comparator`s into a list and pass it into `compare_subs` which will fold on the list to create an overall `Constr.t`.

A new test has been added called `switch_cases_diff_ret` which is very similar to our current `switch_cases` test. In this test, wp should determine that there are two refuted goals: `rax` having different final values, and `deplay_payload` being called in the modified binary.


wp/lib/bap_wp/src/compare.ml
```diff
@@ -113,62 +113,65 @@ let compare_blocks
...
+  postcond, hyps
 
-let compare_subs_empty_post
-    ~input:(_ : Var.Set.t)
-    ~original:(original : Sub.t * Env.t)
-    ~modified:(modified : Sub.t * Env.t)
-    ~smtlib_post:(smtlib_post : string)
-    ~smtlib_hyp:(smtlib_hyp : string)
-  : Constr.t * Env.t * Env.t =
+let compare_subs_empty_post () : comparator * comparator =
```
codyroux, line 174: Why do all these functions return a pair of `comparator`s?

fortunac: They return comparators so that I could have a list of comparators to pass to `compare_subs` that contain all the hypotheses and postconditions. Would there have been a better way to do this?

wp/lib/bap_wp/src/compare.ml
```diff
@@ -113,62 +113,65 @@ let compare_blocks
...
   let goal = Constr.mk_clause [hyps] [pre_combined] in
   goal, env1, env2
 
-let compare_subs_empty
-    ~original:(original : Sub.t * Env.t)
-    ~modified:(modified : Sub.t * Env.t)
-    ~smtlib_post:(smtlib_post : string)
-    ~smtlib_hyp:(smtlib_hyp : string)
-  : Constr.t * Env.t * Env.t =
+let compare_subs_empty () : comparator * comparator =
```
codyroux, line 166: Why does this take unit as input?

fortunac: No particular reason. I can remove it.

wp/plugin/wp.ml
```diff
@@ -156,6 +156,67 @@ let analyze_proj (ctx : Z3.context) (var_gen : Env.var_gen) (proj : project)
...
+let smtlib
+    ~post_cond:(post_cond : string)
+    ~pre_cond:(pre_cond : string)
+  : (Comp.comparator * Comp.comparator) option =
+  if String.is_empty post_cond && String.is_empty pre_cond then
+    None
+  else
+    Some (Comp.compare_subs_smtlib ~smtlib_post:post_cond ~smtlib_hyp:pre_cond)
+
+let sp (arch : Arch.t) : (Comp.comparator * Comp.comparator) option =
```
codyroux, line 192: This feels like it should fail more loudly.

wp/lib/bap_wp/src/compare.mli
```diff
@@ -30,74 +34,93 @@ module Env = Environment
...
-  -> smtlib_post:string
-  -> smtlib_hyp:string
-  -> Constr.t * Env.t * Env.t
+val compare_subs_empty_post : unit -> comparator * comparator
+
+(** Compare two subs by composition for an empty postcondition:
+
+    Give two subroutines and environments, return a postcondition
+    and hypothesis that, when passed to [compare_subs], will generate a
+    precondition which is provable only if (module soundness bugs) the VCs generated
```
codyroux, line 98: module -> modulo

wp/plugin/wp.ml
fortunac: I can add an error message if the architecture isn't x86_64.

wp/lib/bap_wp/src/compare.mli
codyroux: Really everywhere. That's probably my typo, sorry.

codyroux: It might be worth elaborating on this comment, I'm honestly not sure what this comparator does, myself.

wp/lib/bap_wp/src/compare.ml
codyroux: Yea, unit is only useful to delay computation, if something non-trivial is happening in the body, or some side-effect occurs. Neither is the case here.

wp/plugin/wp.ml
codyroux: Seems reasonable.

codyroux:
Modulo my minor comments, this looks great!

wp/lib/bap_wp/src/compare.ml
codyroux: It's fine. I wonder if it would be worth separating the types of comparators that create hypotheses and ones that create postconditions. But honestly if we don't need to, then it's fine.

Issue #160, opened 2020-06-09
Option to remove the printing of the precondition in WP plugin


codyroux:
It's not serving much purpose at this point.

fortunac:
Closed with #168.

PR #162, opened 2020-06-10
Added flag to suppress precondition.

DieracDelta:
Addresses issue #160 by adding in the flag `--wp-suppress-precond`. This flag defaults to false, and when set, prints the precondition to the log instead of stdout in the single binary case (in the comparative case, since the precondition is not printed, the flag has no effect). An example invocation might be:

`bap main --pass=wp --wp-suppress-precond`

### DIFF
--- a/wp/plugin/README.md
+++ b/wp/plugin/README.md
@@ -366,6 +366,11 @@ The various options are:
   contraint that if a memory read or write dereferences a non-null address in the
   original binary, then that same address is also non-null in the modified binary.
   Defaults to false.
+  
+-- `--wp-suppress-precond=[true|false]` If present, suppresses the printing of 
+  generated precondition(s) (shows in the log instead of stdout). Only applies 
+  in the single binary analysis case. In the comparative case, has no effect. 
+  Defaults to false.
 
 ## C checking API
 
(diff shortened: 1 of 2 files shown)


codyroux:
It's worth trying to merge this request with #165, since they have similar purposes and UIs.

One idea might be to have a flag of the sort `--wp-print-constraint=foo` where `foo` is one of `none |  smt-lib | internal`.

The other issue is where to print/not print, since the precondition is printed directly at `wp.ml` whereas `smt-lib` is being printed deeper. Not that big of a deal, but if there's a graceful way to have them happen at the same spot, great.

philzook58:
Maybe I'm crazy, but it strikes me a bit odd to have a flag for "suppressing" the precondition printing.
Does it not make more sense to have a flag --wp-print-precond that defaults to false?
It's the same thing in the end of the day I suppose.

codyroux:
Sometimes you'll want to have a "suppress" flag for something that should be on by default. in this case however, I agree with @philzook58, the default behavior should guide the name: `--wp-print-precond` rather than the other one.

Moot of course if we unify the two flags.

DieracDelta:
This has been resolved in PR #168. Closing!

Issue #165, opened 2020-06-10
Output smtlib2 flag


philzook58:
Dumping the smtlib2 query is useful both pedagogically and for debugging purposes.  
in precondition.ml https://github.com/draperlaboratory/cbat_tools/blob/112ca768925d691fd8781ca873c8cf47a2c641b9/wp/lib/bap_wp/src/precondition.ml#L1069 adding Solver.to_string outputs what you want basically, but we should hide this behind a flag.

``` ocaml
let check ?refute:(refute = true) (solver : Solver.solver) (ctx : Z3.context)
    (pre : Constr.t) : Solver.status =
  printf "Evaluating precondition.\n%!";
  let pre' = Constr.eval pre ctx in
  printf "Checking precondition with Z3.\n%!";
  let is_correct =
    if refute then
      Bool.mk_implies ctx pre' (Bool.mk_false ctx)
    else
      pre'
  in
  Z3.Solver.add solver [is_correct];
  (* Printf.printf "Z3 : %s \n %!" (Z3.Expr.to_string (Z3.Expr.simplify is_correct None)); *)
  Printf.printf "Z3 : %s \n %!" (Z3.Solver.to_string solver);
  Printf.printf "Assertions: %d \n %!" (Z3.Solver.get_num_assertions solver) ;
  Z3.Solver.check solver []
```

gltrost:
Closed with #168 .

PR #168, opened 2020-06-15
Added flag to print smtlib2 (and now also incorporates fix for issue #162)

gltrost:
A new wp flag is created for wp, following the suggestions from philzook58. The flag is:

  --wp-output-smtlib2

with the description:

 _If set, Z3's SMT-LIB 2 query will be dumped, printing the query of the solver. This flag is mainly for pedagogical and debugging purposes. If the flag is not called, it defaults to false and the query won't be printed._

### DIFF
--- a/wp/plugin/wp.ml
+++ b/wp/plugin/wp.ml
@@ -37,7 +37,8 @@ type flags =
     print_path : bool;
     use_fun_input_regs : bool;
     mem_offset : bool;
-    check_null_deref : bool
+    check_null_deref : bool;
+    output_smtlib2 : bool 
   }
 
 let missing_func_msg (func : string) : string =
@@ -273,7 +274,7 @@ let main (flags : flags) (proj : project) : unit =
     else
       analyze_proj ctx var_gen proj flags
   in
-  let result = Pre.check solver ctx pre in
+  let result = Pre.check solver ctx pre ~output_smtlib2:flags.output_smtlib2 in
   let () = match flags.gdb_filename with
     | None -> ()
     | Some f ->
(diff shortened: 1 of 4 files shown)


philzook58:
Addresses #165, but also has been suggested that it be coordinated in some way with #162

wp/plugin/wp.ml
```diff
@@ -152,8 +153,11 @@ let analyze_proj (ctx : Z3.context) (var_gen : Env.var_gen) (proj : project)
   let pre, env = Pre.visit_sub env post main_sub in
   let pre = Constr.mk_clause [Z3_utils.mk_smtlib2_single env flags.pre_cond] [pre] in
   let pre = Constr.mk_clause hyps [pre] in
-  Format.printf "\nSub:\n%s\nPre:\n%a\n%!"
-    (Sub.to_string main_sub) Constr.pp_constr pre;
+  (* Print statement for constrain-style prover output: *)
```
fortunac, line 156: I think we can move this to `Pre.check` so we can handle the print_constr flag in one place by iterating over the list there. The constraint should be the same in both locations.

fortunac: Bonus of being able to print out the `Constr.t` type for both single and comparative analysis as well.

gltrost: The print call on line 160 in wp.ml depends on the value `main_sub`, which is defined in analyze_proj. I can change analyze_proj to output `main_sub` so that it can be called in Pre.check if you feel that isn't too messy.

fortunac: What if we move the printing of `pre` into `Pre.check` and leave the printing of `main_sub` here? It could mirror how `compare_projs` also prints the subroutines. https://github.com/draperlaboratory/cbat_tools/blob/6f6db38da139d5514b9b8ad7e2765a273ef6b6c9/wp/plugin/wp.ml#L258

gltrost: Great idea! I have moved the printing of `pre` into `Pre.check`, and in doing so, when `Pre.check` is called in `wp.ml`, it prints either the pre from `compare_projs` or `analyze_proj`, allowing us to print `pre` for comparative analysis.

fortunac: Looks super good! Do you mind removing the print for `pre` from `wp.ml` so we don't have this code in two places? It looks like we are printing out `pre` twice when I set the flag to `internal`. I think we can leave a simple `printf` for the `main_sub` here.

gltrost: I didn't catch that. 
I'm on it.

PR #172, opened 2020-06-16
Phil/eval speedup

philzook58:
- Changed mk_simple_solver over to mk_solver which may improve z3 solve time
- Early evaluation in get_refuted_goals vastly speeds it up
- Using Boolean.is_true vs Solver.check also speeds up get_refuted_goals

All told brings n-queens example down from 1 minute to 3s.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -233,43 +233,29 @@ let get_refuted_goals ?filter_out:(filter_out = []) (constr : t)
     | Goal g ->
       let goal_val = Expr.substitute g.goal_val olds news in
       let goal_res = eval_model_exn model goal_val in
-      begin
-        match Solver.check solver [goal_res] with
-        | Solver.SATISFIABLE -> Seq.empty
-        | Solver.UNSATISFIABLE ->
-          Seq.singleton
+      if (Z3.Boolean.is_true goal_res)
+      then Seq.empty
+      else Seq.singleton
             { goal = { g with goal_val = goal_val };
               path = current_path;
               reg_map = current_registers }
-        | Solver.UNKNOWN ->
-          failwith (Format.sprintf "get_refuted_goals: Unable to resolve %s" g.goal_name)
-      end
     | ITE (jmp, cond, c1, c2) ->
       let cond_val = Expr.substitute cond olds news in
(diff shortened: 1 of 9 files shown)


fortunac:
This speedup is super nice! 🎉 I think there might be something funky happening in `get_refuted_goals` though. I can't seem to get
```
deploy_payload not called in modified:
    Z3 Expression: false
```
as a refuted goal anymore.

philzook58:
Here is some rationalization of being able to eval_model early.

`get_refuted_goals` is going through the program in the ordinary order, top to bottom. It is a somewhat roundabout straightforward intepreter based on an initial state given by the model. One could take all the registers and memory out of the model and place them in the "olds/news" lists at the beginning of the pass. Then upon encountering a Subst node, these values would be substituted into the right hand side of the Subst. I believe that placing eval_model at this location  in the intepreter achieves the same effect, in a sort of lazy way. It only pulls the value out of the model when it appears on the rhs of a Subst. If that variable has been Subst to something else, it no longer has the value it had at the beginnign of the execution. However, it will then appear in the olds list and be substituted out by the Expr.substitute before eval_model has a chance to get to it. So I think there is some kind of inductive argument these two ways of doing things are the same.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -233,51 +233,48 @@ let get_refuted_goals ?filter_out:(filter_out = []) (constr : t)
...
+          else failwith "Unexpected Expression processing Clause"
       in
       if hyps_false then
         Seq.empty
       else
         List.fold concs ~init:Seq.empty ~f:(fun accum c ->
             Seq.append (worker c current_path current_registers olds news) accum)
     | Subst (e, o, n) ->
-      let n' = List.map n ~f:(fun x -> Expr.substitute x olds news) in
+      let n' = List.map n ~f:(fun x -> Expr.substitute x olds news |> eval_model_exn model) in
```
codyroux, line 277: What does this enable?

codyroux: I guess it's related to your above comment. But still walking through the state of these terms before and after the call to `eval_model` would be useful.

codyroux:
All right, I'm sold, but let's try to keep a close eye and make sure nothing funky occurs when printing models.

PR #178, opened 2020-06-19
Remove default behavior of comparing RAX

### LINKED ISSUE #175: Remove default behavior of comparing RAX
By default, WP will compare the final values of RAX at the end of a subroutine's execution. We should remove this behavior as a default, and only compare these registers when the use specifies `--wp-compare-final-reg-values=RAX`. We should also change the name of the flag from `output-vars` to `compare-final-reg-values`.

fortunac:
Fixes #175. Removes the default behavior of comparing RAX when performing a comparative analysis. The user has to specify which registers' final values they want to compare with `--wp-compare-final-reg-values=<list>`.


wp/plugin/wp.ml
```diff
@@ -332,7 +333,8 @@ module Cmdline = struct
       ~doc:"Amount of times to unroll each loop. By default, wp will unroll each \
             loop 5 times."
 
-  let output_vars = param (list string) "output-vars" ~default:["RAX"; "EAX"]
+  let compare_final_reg_values = param (list string) "compare-final-reg-values"
```
codyroux, line 336: Why "final" rather than "output"?

wp/plugin/wp.ml
```diff
@@ -332,7 +333,8 @@ module Cmdline = struct
...
+      ~default:[]
       ~doc:"List of output variables to compare separated by `,' given the same \
```
codyroux, line 338: Is this docstring still accurate?

codyroux:
The code looks reasonable, but I worry a bit about the user experience: an amateur user might get confused about having to come up with register names for the target architecture on their own. Is there any way to provide some sensible defaults with the flag activated?

I am all for not doing anything by default without *any* flags, but probably something reasonable should happen with `--wp-compare-final-regs` and no additional info.

wp/plugin/wp.ml
fortunac: I think the word final is used because the registers that are being compared aren't necessarily going to be the output registers, but will be the final value of any register.

fortunac:
> The code looks reasonable, but I worry a bit about the user experience: an amateur user might get confused about having to come up with register names for the target architecture on their own. Is there any way to provide some sensible defaults with the flag activated?
> 
> I am all for not doing anything by default without _any_ flags, but probably something reasonable should happen with `--wp-compare-final-regs` and no additional info.

This is very true. If the user doesn't have RAX specified, nothing is going to happen. I have no complaints for having RAX be the default comparison register. I think @jtpaasch might have some good opinions as well.

wp/plugin/wp.ml
fortunac: It is not. I can update this.

jtpaasch:
I'm trying to imagine the user that Cody is thinking of, and I'm trying to see what they would expect here. Suppose I'm a user who understands very little about binary programs. I do understand that there are registers, and that subroutines put their outputs in some of those registers. However, I do not understand any specific ABI, and so I don't really know *which* registers subroutines are supposed to put their output in. This seems to be the user Cody has in mind (i.e., the user who doesn't know enough to specify *which* registers they should put as values for this flag). 

As this beginner, since I don't really understand which registers matter, if this flag has a default behavior, then I would expect the default behavior to be this: `wp` will check *all* registers that the subroutines I'm comparing put output in (whatever those registers may be). As a user, if you're giving me a default here, I think that means I don't have to worry about which registers get affected, because you handle that for me. So I'm expecting you to check all the registers that matter.

If it is possible and feasible (from a technical point of view) for `wp` to figure out which registers the subroutines at hand will put their output in, then I think the default behavior of this flag should be to check all of those. If that is exactly RAX, then great, I think that default makes sense.

On the other hand, if RAX is not the only register that may be relevant, and if it's not possible or feasible (from a technical point of view) for `wp` to figure out which registers are affected and check all of them, then I think it's misleading to me, the beginner, to have a default of RAX here. In fact, I think it's misleading to have any default here. By having RAX (or anything else) as the default, you're giving me the impression that RAX (or whatever else it is) is precisely the register(s) that matters.

Hmm. I don't know the calling convention here. Is RAX really just a good default *guess*? If so, then I don't think it should be a default. I think in that case it would be more intuitive to me as a beginner to have to specify a value myself for this flag (so there's no default), but I'd like to see RAX suggested in the documentation as a good first thing to try. 

For example, I think it would make sense to see in the documentation something like this: "Specifying `--compare-final-regs=RAX,RDI,R9` tells `wp` to compare the values that the subroutines you're comparing can put in the registers `RAX`, `RDI`, and `R9` after they finish running. If it's possible for the two subroutines to put different values in these registers after they finish running, then `wp` will find this, and show you an example of how to cause it. If you do not know which registers to compare, try `RAX` as a first pass. In `x86_64`, subroutines typically put at least one of their outputs in RAX."

fortunac:
BAP has a way of determining which registers the output value is stored in, so it is definitely possible to use that as a way to set a default register for the flag. It is usually RAX for x86_64, but going through BAP's arg terms is better than just hardcoding in RAX. 

I can see two possibilities:
1. no default value for the flag, and the suggestion in the documentation is based off of the register(s) that BAP finds as the output.
2. the register(s) BAP finds as output is set as the default for the flag. It will usually be RAX, but could possibly be different for other architectures/ABIs. For example, I believe the register for 32-bit ARM is r0.

I am slightly leaning towards option 1. That way, there is no funny magic happening behind the scenes that affects the result of the analysis for the user.

jtpaasch:
Can I use this flag to have `wp` check other registers, beyond those that are formally considered "output" registers? I mean, yeah, there's the arg term for a subroutine that specifies which register is the "output" register. But can I use this `--compare-final-regs` flag to check other things?

fortunac:
Yes!

jtpaasch:
So this flag isn't really about comparing the "output" register. It's about comparing any arbitrary set of registers at the end of a function?

fortunac:
The initial use case of the flag was for comparing the output registers when we first wrote it, but you can actually specify any register as long as it exists in the subroutine. It definitely makes more sense to say it compares the final state of any arbitrary set of registers at the end of the subroutine now.

jtpaasch:
If that's how we want to message the purpose of this flag, then yeah, I wouldn't want any default behavior for this. I agree that would lead to too much magic funny business there. My vote is that the user should say which registers they care to compare.

wp/plugin/wp.ml
codyroux: How about `compare_post_reg_values`?

codyroux:
2 Comments: I'm fine with having no defaults, as long as the doc string gives plenty of examples. Second, what's the status on this? Are we ok with everything else? I am personally.

fortunac:
I still need to get the docs to give examples of registers to give to the flag depending on the architecture. Haven't gotten to this yet.

wp/plugin/wp.ml
fortunac: I have no problems with `compare_post_reg_values`.

PR #179, opened 2020-06-23
Add functionality to assert register value in model for plugin tests

### LINKED ISSUE #88: Add more checks to WP unit tests
Currently in each of the unit tests, we are only checking to see if Z3 returns `SAT` or `UNSAT`:

https://github.com/draperlaboratory/cbat_tools/blob/0b9e9737eca958ab3dc5176a7cfec38c91c910c4/wp/lib/bap_wp/tests/unit/test_precondition.ml#L31-L41

https://github.com/draperlaboratory/cbat_tools/blob/0b9e9737eca958ab3dc5176a7cfec38c91c910c4/wp/plugin/tests/test_wp.ml#L26-L28

In the `switch_cases` test, `SAT` was returned as expected, but the generated model specified `RDI = 0` where `RDI = 3` was expected. We should add more assertions to our test cases to prevent this:

- Register values:
  - obtain the Z3 model after running check
  - obtain the value of a register
  - use `oUnit`'s `assert_equals` to compare the register value with the expected value
- Possibly paths?
  - is it feasible/possible to check if we are going down the correct path?

fortunac:
Fixes #88. If you want to assert the value of a register in the model, it is now possible to pass in a list of registers and their expected values to `test_plugin`.

### DIFF
--- a/wp/plugin/tests/test_wp.ml
+++ b/wp/plugin/tests/test_wp.ml
@@ -39,12 +39,15 @@ let check_z3_result (line : string) (expected : string) (ctxt : test_ctxt) : uni
   else
     ()
 
+(* Checks to see if a line of output belongs to the countermodel where a register
+   is given a value. If the register is one that we have an expected value for,
+   assert the model's value is equal to our expected value. *)
 let check_model (line : string) (regs : (string * string) list) (ctxt : test_ctxt)
   : unit =
   List.iter regs ~f:(fun (reg, expected) ->
       (* Our model is printed in the form "reg |-> value". This regex will break
          if this changes. *)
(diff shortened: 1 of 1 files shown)


wp/plugin/tests/test_wp.ml
```diff
@@ -29,22 +29,46 @@ let unsat : string = "UNSAT!"
...
+  if List.exists [sat; unsat; unknown] ~f:(String.equal line) then
+    assert_equal
+      ~ctxt
+      ~printer:String.to_string
+      ~cmp:String.equal
+      expected line
+  else
+    ()
+
+let check_model (line : string) (regs : (string * string) list) (ctxt : test_ctxt)
```
codyroux, line 42: Can we have comments for these?

codyroux:
This looks great, I love it.

PR #183, opened 2020-06-26
Gltrost/debug flags

gltrost:
My code adds a debug flag with four options:

- `solver` : shows various stats for `solver` in the function `main` in `wp.ml`

- `global` : calls ` Z3.set_global_param "verbose" "10" `, also in the function `main`

- `constr.t` : gives statistics on `pre` in the function `main` by calling `Constr.get_stats`. One area for improvement with this function is that it is very slow to run. 

- `eval` : gives statistics on the list values `olds` and `news` in `eval_aux` when called in the function `eval` in `constraint.ml`. List includes the mean, max and standard deviation of the list sizes.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.mli
+++ b/wp/lib/bap_wp/src/constraint.mli
@@ -100,8 +100,10 @@ val mk_ite : Bap.Std.Jmp.t -> z3_expr -> t -> t -> t
 val mk_clause : t list -> t list -> t
 
 (** [eval c ctx] evaluates the constraint [c] to a Z3 expression,
-    using the standard Z3 encoding of if-then-else and clauses. *)
-val eval : ?debug:(string list) -> t -> Z3.context -> z3_expr
+    using the standard Z3 encoding of if-then-else and clauses.
+    If ~debug is called and is set to true, then statistics for
+    eval will be printed. *)
+val eval : ?debug:(bool) -> t -> Z3.context -> z3_expr
 
 (** [substitute c [e1;...;e_n] [d1;...;d_n]] substitutes each
     occurence in [c] of the Z3 expression [ei] with the expression
@@ -130,8 +132,17 @@ val eval_model_exn : Z3.Model.model -> z3_expr -> z3_expr
 val get_model_exn : Z3.Solver.solver -> Z3.Model.model
 
 
+(** stats, get_stats and print_stats are used for debugging purposes when
+    the flag --wp-debug=constraint-stats is set. **)
+
+(** Contains statistics for values of type t, keeping track of number of goals,
+    ites, clauses and subs that compose t **)
 type stats
 
+(** [get_stats t stats] collects statistics of t, so that the statistics are
+    held in stats **)
 val get_stats : t -> stats -> unit
 
+(** [print_stats t] calls get_stats to collect and print the statistics
+    for the number of goals, ites, clauses and subs that compose t **)
 val print_stats : t -> unit
--- a/wp/lib/bap_wp/src/constraint.mli
+++ b/wp/lib/bap_wp/src/constraint.mli
@@ -101,7 +101,7 @@ val mk_clause : t list -> t list -> t
 
 (** [eval c ctx] evaluates the constraint [c] to a Z3 expression,
     using the standard Z3 encoding of if-then-else and clauses. *)
-val eval : t -> Z3.context -> z3_expr
+val eval : ?debug:(string list) -> t -> Z3.context -> z3_expr
 
 (** [substitute c [e1;...;e_n] [d1;...;d_n]] substitutes each
     occurence in [c] of the Z3 expression [ei] with the expression
--- a/wp/lib/bap_wp/src/constraint.mli
+++ b/wp/lib/bap_wp/src/constraint.mli
@@ -128,3 +128,10 @@ val eval_model_exn : Z3.Model.model -> z3_expr -> z3_expr
     if retrieving the model fails in the case that [check] was not yet called
     to the solver or the result of the last [check] did not result in SAT. *)
 val get_model_exn : Z3.Solver.solver -> Z3.Model.model
+
+
+type stats
+
+val get_stats : t -> stats -> unit
+
+val print_stats : t -> unit
--- a/wp/lib/bap_wp/src/precondition.mli
+++ b/wp/lib/bap_wp/src/precondition.mli
@@ -274,11 +274,13 @@ val visit_block : Env.t -> Constr.t -> Bap.Std.Blk.t -> Constr.t * Env.t
 val visit_sub : Env.t -> Constr.t -> Bap.Std.Sub.t -> Constr.t * Env.t
 
 (** Calls Z3 to check for a countermodel for the precondition of a BIR program. If
-    refute is set to false, it checks for a model instead. *)
+    refute is set to false, it checks for a model instead. If ~print_constr is
+    called and contains "smtlib", then Z3's solver (smt lib 2) will be printed.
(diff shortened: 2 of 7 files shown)


wp/lib/bap_wp/src/constraint.mli
```diff
@@ -101,7 +101,7 @@ val mk_clause : t list -> t list -> t
 
 (** [eval c ctx] evaluates the constraint [c] to a Z3 expression,
     using the standard Z3 encoding of if-then-else and clauses. *)
-val eval : t -> Z3.context -> z3_expr
+val eval : ?debug:(string list) -> t -> Z3.context -> z3_expr
```
fortunac, line 104: Are the available debug options `"solver", "global", "constr.t", "eval"`? Can we add a quick comment about that here?

wp/lib/bap_wp/src/constraint.mli
```diff
@@ -128,3 +128,10 @@ val eval_model_exn : Z3.Model.model -> z3_expr -> z3_expr
     if retrieving the model fails in the case that [check] was not yet called
     to the solver or the result of the last [check] did not result in SAT. *)
 val get_model_exn : Z3.Solver.solver -> Z3.Model.model
+
+
+type stats
+
+val get_stats : t -> stats -> unit
+
+val print_stats : t -> unit
```
fortunac, line 137: Can we add a few comments about `stats`, `get_stats`, and `print_stats` here? What kind of stats is the user expecting?

wp/lib/bap_wp/src/constraint.mli
gltrost: Those are the four options.
eval can use the "eval" option, if set, for printing certain statistics.
I'll add that to the comments.

fortunac: Thanks!

wp/lib/bap_wp/src/constraint.mli
gltrost: I'm adding these now.

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -278,6 +278,7 @@ val visit_sub : Env.t -> Constr.t -> Bap.Std.Sub.t -> Constr.t * Env.t
 val check
   : ?refute:bool
   -> ?print_constr: (string list)
+  -> ?debug: (string list)
```
fortunac, line 281: Does `debug` take the same options as `eval`? While we're here, can we also add the `print_constr` options to the doc as well?

philzook58:
I found the naming of the flags a little confusing.

I'd suggest maybe
z3-stats 
z3-verbose
constraint-stats
eval-constraint-stats

But I'm open to suggestions

wp/lib/bap_wp/src/precondition.mli
gltrost: I knew you would catch that one next! Already done!

fortunac:
> I found the naming of the flags a little confusing.
> 
> I'd suggest maybe
> z3-stats
> z3-verbose
> constraint-stats
> eval-constraint-stats
> 
> But I'm open to suggestions

I am a fan of the `z3-verbose` name. I feel like the `z3-stats` flag should have the word `solver` in it somewhere because it has to do with Z3's solver specifically.

gltrost:
> > I found the naming of the flags a little confusing.
> > I'd suggest maybe
> > z3-stats
> > z3-verbose
> > constraint-stats
> > eval-constraint-stats
> > But I'm open to suggestions
> 
> I am a fan of the `z3-verbose` name. I feel like the `z3-stats` flag should have the word `solver` in it somewhere because it has to do with Z3's solver specifically.

I think these are good ideas. How about `z3-solver-stats` to be more precise?

fortunac:
The stats being printed out look nice! I do have a question about where we should be printing them. Would it be best to have these printed to STDOUT or to the logs?

wp/plugin/README.md
```diff
@@ -373,6 +373,12 @@ The various options are:
   what is stated. Both can also be called like `--wp-print-constr=internal,smtlib`.
   If the flag is not called, it defaults to printing neither.
 
+- `--wp-debug=[solver|global|constr.t|eval] or some combination`. If set, debug
```
philzook58, line 376: Maybe put "some comma delimited combination"

philzook58:
I'm personally fine with them being printed. But that's because i can never remember where logs go.

wp/lib/bap_wp/src/precondition.mli
gltrost: > I'm personally fine with them being printed. But that's because i can never remember where logs go.

I'm fine with either as well.

gltrost:
> I'm personally fine with them being printed. But that's because i can never remember where logs go.

I'm fine with either as well.

wp/lib/bap_wp/src/precondition.mli
gltrost: A question on comments. Since the debug-flags, by design, break the black-box nature of functions they are called in, can the comments concerning the debug flags also be allowed to talk about the function's inner implementation, or should they try to keep things opaque to respect the interface.

fortunac:
> > I'm personally fine with them being printed. But that's because i can never remember where logs go.
> 
> I'm fine with either as well.

I guess we can keep them as stdout, and if it get's problematic, we can move it to the logs later if needed.

wp/lib/bap_wp/src/precondition.mli
fortunac: I would say go ahead, but there is no need to get too detailed as long as you're able to mention the options and what each one prints out.

gltrost:
Also, before anyone else mentions it, the entire `debug` flag is passed into `Pre.check` in `wp.ml`, which then passes in the entire flag to `Constr.eval` inside `check` in `precondition.ml`, so that the flag is then finally cased on within `eval` in `Constraint.ml`.

If you'd like, I can make it so `Pre.check` and `Constr.eval` are only given a bool instead of the whole debug flag as an input.

fortunac:
> Also, before anyone else mentions it, the entire `debug` flag is passed into `Pre.check` in `wp.ml`, which then passes in the entire flag to `Constr.eval` inside `check` in `precondition.ml`, so that the flag is then finally cased on within `eval` in `Constraint.ml`.
> 
> If you'd like, I can make it so `Pre.check` and `Constr.eval` are only given a bool instead of the whole debug flag as an input.

I actually think this is a great idea! We can check for the specific flag in `wp.ml` and just pass in the bool as necessary. We won't have to maintain all those docs with the different options in the mlis this way. Sorry for the extra work.

wp/plugin/README.md
```diff
@@ -373,6 +373,14 @@ The various options are:
   what is stated. Both can also be called like `--wp-print-constr=internal,smtlib`.
   If the flag is not called, it defaults to printing neither.
 
+- `--wp-debug=[z3-solver-stats|z3-verbose|constrain-stats|eval-constraint-stats]
```
fortunac, line 376: Is `constrain-stats` a typo? Should it be `constraint-stats`?

fortunac:
The changes look good! A small nitpick is that we renamed one of the flags to `z3-verbose`, but the docs refer to it as a global parameter. This might be confusing for a user, and we can probably abstract that out to just say we are changing Z3's verbosity level.

gltrost:
> we renamed one of the flags to `z3-verbose`

That's a good nit-pick. I just pushed the changes for that.

fortunac:
Just that one typo, and I think this is good to go! Does @philzook58 have any more comments?

philzook58:
Looks good to me

PR #184, opened 2020-06-29
Phil eval speedup2

philzook58:
The size of the olds/news list occurring in the Constr.eval function seems to be a major source of slowdown. One step towards correcting this is to use a map data structure. Because without looking deeper under the hood the Z3 ocaml api only accepts lists, It appears to be most efficient to still maintain olds/news as lists for the moment, but we can prune multiple occurrences of the same keys in the olds list. In addition, it appears that moving the most recently changed variables to the head of the olds list speeds things up by making it faster to find commonly used variables. One problem is that bap is generating virtual memory addresses which increase the sizes of the lists but cannot be easily garbage collected, despite being accessed very rarely (often only once or twice). 
I believe every time Expr.substitute is called the olds/news list are traversed and unpacked into an array for z3 internal use.
The list processing and `Expr.substitute` functions are still the bottleneck in the slow examples I've looked at.

Possibilities for further improvements:
- The `new` expression being processed in `Subst(c, old, new)` rarely contains more than a few variables. It is possible to traverse the expression and fill olds/news only with the variables needed for. This allows the use of efficient Map data structures. Doing this naively was found to be more complicated, of questionable correctness, and not much faster. Now instead of lots of time being spent in Expr.substititue a lot of time was spent in the variable collection pass. I think this could be made more efficient by peeking under the z3 api to the z3native module. The Map structures were not all that fast either compared to the simple most recently accessed element of list strategy discussed above as Expr.compare is not super cheap or too often used.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -235,18 +235,27 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
       let hyps_expr = eval_conjunction hyps in
       Bool.mk_implies ctx hyps_expr concs_expr
   | Subst (c, o, n) ->
+    (* The following should be functionally equivalent to [eval_aux ~stats:stats c (olds @ o) (news @ n') ctx] 
+       with changes made to achieve better performance. Duplicate keys are pruned from the olds list.
+       
+       NOTE : It is assumed as an invariant that the olds list does not contain duplicate keys. Failure to maintain this invariant
+       may result in incorrect behavior.
+
+       Using a Map data structure was tried and it was found that shuffling in and out to lists for Expr.substitute made it slow. 
+       Instead, two lists are maintained and pruned of duplicate keys. The most recently used items are put at the front of the list 
+       to make it fast to find the most commonly used variables.
+       *)
     let n' = List.map n ~f:(fun x -> Expr.substitute x olds news) in
     if (List.length n') = 1 (* It appears we generate overwhelmingly size 1 lists *)
       then 
       begin
         let o = List.hd_exn o in
         let n' = List.hd_exn n' in
-        (* A Map was tried. It was found that shuffling in and out to lists for Expr.substitute made it not worth it  *)
-        (* The idea here is that popping items most recently used to the front of the list makes it fast to find the most commonly used variables *)
-        (* It is assumed as an invariant that the olds list does not contain duplicate keys *)
         let findn e l =
+            (* [findn' n e l] finds element [e] in list [l] while maintainting an accumulator of the current index [n] *)
             let rec findn' n e l = match l with | x :: xs ->  if Expr.equal x e then Some n else findn' (n + 1) e xs | [] -> None in
             findn' 0 e l in
+        (* [removen_exn n l] removes item at position [n] in list [l] *)
         let rec removen_exn n l = match l with | x :: xs ->  if (n = 0) then xs else x :: (removen_exn (n - 1) xs) |  [] -> failwith "Improper removen use" in
         let oind = findn o olds in
         match oind with
@@ -256,6 +265,7 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
                         eval_aux ~stats:stats c olds news' ctx
       end
     else
+      (* Substitutions larger than 1 variable long uses zipping and less efficient assoclist facilities to delete duplicates *)
       let oldsnews = List.zip_exn olds news in
       let o', n' =  List.fold2_exn o n' ~init:oldsnews ~f:(fun on' old new' -> List.Assoc.add on' ~equal:Expr.equal old new') |> List.unzip   in 
       eval_aux ~stats:stats c o' n' ctx 
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -359,15 +359,9 @@ let get_refuted_goals ?filter_out:(filter_out = []) (constr : t)
         worker c2 current_path current_registers olds news
       else
         failwith (Format.sprintf "get_refuted_goals: Unable to resolve branch \
-<<<<<<< HEAD
                                     condition %s at %s"
                                     (Z3.Expr.to_string cond_res)
                                     (jmp |> Term.tid |> Tid.to_string))
-=======
-                                  condition %s at %s" 
-                    (Z3.Expr.to_string cond_res)
-                    (jmp |> Term.tid |> Tid.to_string))
->>>>>>> ed3b7c6... Changed indenting and added fast path to get_refuted_goals subst
     | Clause (hyps, concs) ->
       let hyps_false =
         let hyp_vals =
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -235,9 +235,7 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
       let hyps_expr = eval_conjunction hyps in
       Bool.mk_implies ctx hyps_expr concs_expr
   | Subst (c, o, n) ->
-    let n' = List.map n ~f:(fun x -> 
-                                   let[@landmark "oldsnews" ] olds, news = get_vars' x olds news |> List.unzip in
-                                   Expr.substitute x olds news) in
+    let n' = List.map n ~f:(fun x -> Expr.substitute x olds news) in
     if (List.length n') = 1 (* It appears we generate overwhelmingly size 1 lists *)
       then 
       begin
@@ -252,7 +250,7 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
         let rec removen_exn n l = match l with | x :: xs ->  if (n = 0) then xs else x :: (removen_exn (n - 1) xs) |  [] -> failwith "Improper removen use" in
         let oind = findn o olds in
         match oind with
-          | None -> eval_aux c (o :: olds) (n' :: news) ctx
+          | None -> eval_aux ~stats:stats c (o :: olds) (n' :: news) ctx
           | Some ind -> let news' = n' :: (removen_exn ind news) in
                         let olds  = o  :: (removen_exn ind olds) in 
                         eval_aux ~stats:stats c olds news' ctx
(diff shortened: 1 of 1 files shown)


wp/lib/bap_wp/src/constraint.ml
```diff
@@ -209,18 +209,43 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
       (eval_aux ~stats:stats c2 olds news ctx)
   | Clause (hyps, concs) ->
     let eval_conjunction conj =
-      List.map conj ~f:(fun c -> eval_aux ~stats:stats c olds news ctx)
-      |> Bool.mk_and ctx
+      if List.length conj = 1 then 
+        eval_aux ~stats:stats (List.hd_exn conj) olds news ctx (* This is tail recursive. Avoids And node. *)
+      else
+        List.map conj ~f:(fun c -> eval_aux ~stats:stats c olds news ctx)
```
fortunac, line 215: Do you know if there are performance impacts if we create an `and` with an empty list?

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -209,18 +209,43 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
...
+    if (List.length n') = 1 (* It appears we generate overwhelmingly size 1 lists *)
+      then 
+      begin
+        let o = List.hd_exn o in
+        let n' = List.hd_exn n' in
+        (* A Map was tried. It was found that shuffling in and out to lists for Expr.substitute made it not worth it  *)
+        (* The idea here is that popping items most recently used to the front of the list makes it fast to find the most commonly used variables *)
+        (* It is assumed as an invariant that the olds list does not contain duplicate keys *)
+        let findn e l =
+            let rec findn' n e l = match l with | x :: xs ->  if Expr.equal x e then Some n else findn' (n + 1) e xs | [] -> None in
```
fortunac, line 235: What does `n`, `e`, and `l` stand for here?

wp/lib/bap_wp/src/constraint.ml
philzook58: I think there were, but now I am not 100% sure.

wp/lib/bap_wp/src/constraint.ml
philzook58: e is the expression one is searching for in list l. n is a carried accumulator parameter. It is needed to give the index at which one finds e. Every time e isn't found it is incremented.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -209,18 +209,43 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
...
     if List.is_empty hyps then
-      concs_expr
+      eval_conjunction concs(* possibly tail recursive? *)
     else
+      let concs_expr = eval_conjunction concs in
       let hyps_expr = eval_conjunction hyps in
       Bool.mk_implies ctx hyps_expr concs_expr
   | Subst (c, o, n) ->
     let n' = List.map n ~f:(fun x -> Expr.substitute x olds news) in
-    eval_aux ~stats:stats c (olds @ o) (news @ n') ctx
```
codyroux, line 223: Perhaps a silly comment, but can we keep the above code as a comment explaining that that is the notional semantics of the following block of code?

philzook58: That makes sense

wp/lib/bap_wp/src/constraint.ml
fortunac: 👍

wp/lib/bap_wp/src/constraint.ml
fortunac: Do you think we can add this as a comment because I won't remember this after a month?

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -310,7 +345,9 @@ let get_refuted_goals ?filter_out:(filter_out = []) (constr : t)
             Seq.append (worker c current_path current_registers olds news) accum)
     | Subst (e, o, n) ->
       let n' = List.map n ~f:(fun x -> Expr.substitute x olds news |> eval_model_exn model) in
-      worker e current_path current_registers (olds @ o) (news @ n')
+      let oldsnews = List.zip_exn olds news in
```
fortunac, line 348: Do we generate list sizes of 1 a lot like the case in `eval` or does `get_refuted_goals` not need that special case?

philzook58: That's a good point. I think they are using the same Constr.t objects, so I expect also this one to include almost exclusively lists of size 1.

fortunac:
Tests are passing, so I have no problems with these changes. The logic makes sense to me, but having Cody take a closer look will make me feel better. Just a small nitpick, but can you keep the lines of code to about 80-90 characters long and run `ocp-indent` on this?

I'm excited for the speedup! 🎉

wp/lib/bap_wp/src/constraint.ml
philzook58: It appears n-queens-7 goes from ~15s to ~14s by replicating the list logic here. Removing the fancy logic from eval in turn takes the total runtime from 15s to 21.5s

philzook58: A majority of the time is still in the precondition constraint evaluation phase and get_refuted_goals is relatively quick.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -222,18 +222,72 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
...
+            match l with 
+            | x :: xs ->  if Expr.equal x e 
+              then Some n 
+              else findn' (n + 1) e xs 
+            | [] -> None in
+          findn' 0 e l in
+        (* [removen_exn n l] removes item at position [n] in list [l] *)
+        let rec removen_exn n l = 
+          match l with 
+          | x :: xs ->  if (n = 0)
```
codyroux, line 270: (Quibble) I'd have the if on the following line. Did you run ocp-indent?

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -312,9 +366,9 @@ let get_refuted_goals ?filter_out:(filter_out = []) (constr : t)
       let hyps_false =
         let hyp_vals =
           List.map hyps ~f:(fun h -> eval_model_exn model (eval_aux h olds news ctx)) in
-          if (List.for_all hyp_vals ~f:(fun e -> Z3.Boolean.is_true e || Z3.Boolean.is_false e))
-          then not (List.for_all hyp_vals ~f:Z3.Boolean.is_true)
-          else failwith "Unexpected Expression processing Clause"
+        if (List.for_all hyp_vals ~f:(fun e -> Z3.Boolean.is_true e || Z3.Boolean.is_false e))
+        then not (List.for_all hyp_vals ~f:Z3.Boolean.is_true)
```
codyroux, line 370: Can we use a `List.exists ...` instead of a `not (List.forall ...)`? I think it might be a bit clearer.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -222,18 +222,72 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
...
-    eval_aux ~stats:stats c (olds @ o) (news @ n') ctx
+    if (List.length n') = 1 (* It appears we generate overwhelmingly size 1 lists *)
+    then 
+      begin
+        let o = List.hd_exn o in
+        let n' = List.hd_exn n' in
+        let findn e l =
+          (* [findn' n e l] finds element [e] in list [l] while maintaining an 
+             accumulator of the current index [n] *)
+          let rec findn' n e l =
```
codyroux, line 260: Can we just use `List.findi`? Feels like we're re-inventing the wheel.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -323,7 +377,39 @@ let get_refuted_goals ?filter_out:(filter_out = []) (constr : t)
             Seq.append (worker c current_path current_registers olds news) accum)
     | Subst (e, o, n) ->
       let n' = List.map n ~f:(fun x -> Expr.substitute x olds news |> eval_model_exn model) in
-      worker e current_path current_registers (olds @ o) (news @ n')
+      if (List.length n') = 1 (* It appears we generate overwhelmingly size 1 lists *)
```
codyroux, line 380: Looks like there's quite a bit of code duplication here. Is this any different than what's above? Can you re-factor this?

It's gonna be a nightmare to debug if we have to maintain them in parallel.

codyroux:
I'm sorry, but I feel that there are some small, but non-trivial structural issues with the code. As far as I can tell, it's correct, but there's code duplication and I feel like there might be some opportunity to use Core libraries to a fuller extent, though indeed there doesn't seem to be a generic "remove at position`i`" function.

In the interest of maintainability, I'd suggest re-factoring things a bit further.

I can help if you feel frustrated by doing this yourself.

wp/lib/bap_wp/src/constraint.ml
philzook58: Where is such a function

philzook58: Oh I found it. nice.

wp/lib/bap_wp/src/constraint.ml
philzook58: I did. My ocp-indent does not create newlines I don't think

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -305,16 +350,16 @@ let get_refuted_goals ?filter_out:(filter_out = []) (constr : t)
         worker c2 current_path current_registers olds news
       else
         failwith (Format.sprintf "get_refuted_goals: Unable to resolve branch \
-                                    condition %s at %s"
-                                    (Z3.Expr.to_string cond_res)
-                                    (jmp |> Term.tid |> Tid.to_string))
+                                  condition %s at %s"
+                    (Z3.Expr.to_string cond_res)
+                    (jmp |> Term.tid |> Tid.to_string))
```
philzook58, line 355: ocp-ident keeps changing this string. Is this a problem @fortunac ?

fortunac: Nope! Should be fine.

PR #185, opened 2020-06-29
Trip asserts and function spec flags

### LINKED ISSUE #167: Add flag for tripping asserts
By default, WP looks for paths that will reach an assert. However, we should allow the user to ask WP to look for asserts rather than making it the default behavior. We should add the flag `--trip-assert` as a way to ask WP to look for these paths.

The default function specs area defined here:
https://github.com/draperlaboratory/cbat_tools/blob/112ca768925d691fd8781ca873c8cf47a2c641b9/wp/lib/bap_wp/src/precondition.ml#L648

### LINKED ISSUE #176: Goto String test failing
@fortunac says

> With bap 2.1, bap is able to determine arg_t terms for goto_string's subroutine, giving it a different spec than we were using. The output is correct for the new spec, it's just different from the old spec and so I need to figure out which spec we should really be using.

fortunac:
Fixes #167 and #176.

Adds two flags to the wp plugin:
1. `--wp-trip-asserts` which will look for inputs to the subroutine that would cause an `__assert_fail` or `__VERIFIER_error` to be reached
2. `--wp-fun-specs=<list>` that allows the user to pick which specs they want to be enabled for the analysis.
    - the user can specify `--wp-fun-specs=all` if they want to enable all of the specs, which is defined in `Pre.default_fun_specs`

Allowing the user to choose specs fixes #176, as the intended spec of `chaos_caller_saved` will be used in this case.

### DIFF
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -537,6 +537,8 @@ let spec_arg_terms (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
   else
     None
 
+(* If RAX is found on the left hand side of an assignment, this spec will
+   chaos RAX as an output from the target function. *)
 let spec_rax_out (sub : Sub.t) (arch : Arch.t) : Env.fun_spec option =
   (* Calling convention for x86 uses EAX as output register. x86_64 uses RAX. *)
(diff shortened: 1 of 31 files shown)


fortunac:
Closing for now in favor of #192.

PR #187, opened 2020-06-30
Moving bildb to the public repo.

jtpaasch:
Updated the README (so check that out) and added bildb.


ivg:
yay!

jtpaasch:
Gonna merge. Aside from `bildb` being added, the other changes are just to the READMEs.

PR #192, opened 2020-07-07
Fortunac/trip asserts2

### LINKED ISSUE #167: Add flag for tripping asserts

fortunac:
Fixes #167. The user needs to specify `--wp-trip-asserts` in order to find inputs that would reach an assert false.

### DIFF
--- a/wp/README.md
+++ b/wp/README.md
@@ -398,6 +398,10 @@ The various options are:
   eval-constraint-stats respectively. If the flag is not called, it defaults to
   printing none of them.
 
+- `--wp-trip-asserts=[true|false]`. If set, WP will look for inputs to the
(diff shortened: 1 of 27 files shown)


jtpaasch:
Looks good! Let's merge it!

jtpaasch:
Looks good to me.

Issue #193, opened 2020-07-08
Improve output from WP


fortunac:
In the case of a SAT result, WP prints out a lot of input and can overwhelm new users with a huge dump of information.

In the long term, we should make the language of SAT/refuted goals clearer to a user. A user without prior knowledge of SAT solving may not know what SAT or refuted goals mean. A possible message for SAT could be something similar to "Z3 was able to find a set of registers and memory values that result in <the property we are checking>".

For now, we can bundle up the output of refuted goals into a flag called `--wp-show-refuted-goals`, and only print when this flag is set. Currently, refuted goals print out their tagged named, the concrete values based on the model, and the Z3 expression representing the goals. This information is probably more useful to a more advanced user who wants to see the specific details about the refuted goals.

If we do this, we are going to have several flags related to debugging/verbosity, and we may need to somehow bundle this up, but this can be left for a future ticket.

fortunac:
We are starting to have a lot of flags for printing. We can possibly combine them into a single flag, `--wp-show=bir,refuted-goals,paths`.

fortunac:
Potential outputs for refuted goals:
- check calls: `<function>` was not called in the original binary, but has been called in the modified binary.
- register comparisons: `<register>` has different values at the end of the subroutine's execution
- assert fail: An `assert_fail` has been reached
- check null deref: A NULL address was dereferenced in a memory read/write
- check null deref (comparison): A NULL address was dereferenced in a memory read/write in the modified binary that was not dereferenced in the original binary

PR #197, opened 2020-07-08
Add flag for printing refuted goals

fortunac:
Addresses part of #193. Adds a flag, `--wp-print-refuted-goals`. Refuted goals won't be calculated or printed out if the flag isn't set.


jtpaasch:
I give it a thumbs up.

PR #198, opened 2020-07-09
Filter the graph visited by WP by the nodes reachable from the start …

codyroux:
…node. This is required when we use our loop-unfolding callback.

This is the first of several fixes required to make our loop unrolling actually work.

I'll keep pushing to this branch, but we might want to make incremental merges to make the review more palatable.


fortunac:
Makes sense to me and tests are passing!

codyroux:
You wanna do the honors?

PR #200, opened 2020-07-10
Convert register values in model to hex

### LINKED ISSUE #194: Make Z3 model values more uniform
We might want to format the values printed in the model as all hex values. Currently, they all begin with #, which is not exactly standard, and some are b while others are x.

fortunac:
Fixes #194.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -51,11 +51,16 @@ let goal_to_string (g : goal) : string =
   Format.sprintf "%s: %s%!" g.goal_name (Expr.to_string (Expr.simplify g.goal_val None))
 
 let expr_to_hex (exp : z3_expr) : string =
-  let decimal = exp |> BV.numeral_to_string |> Int.of_string in
-  let size = exp |> Expr.get_sort |> BV.get_size |> Int.to_float in
+  let decimal = BV.numeral_to_string exp in
+  let size = exp |> Expr.get_sort |> BV.get_size in
+  let word = Word.of_string (Printf.sprintf "%s:%du" decimal size) in
+  let pp_word ppf =
+    Word.pp_generic ~case:`lower ~prefix:`none ~suffix:`none ~format:`hex ppf in
+  let word_str = Format.asprintf "%a" pp_word word in
   (* Calculate how many 0s to pad the output with. *)
-  let width = size /. 4. |> Float.round_up |> Int.of_float in
-  Format.sprintf "0x%0*x" width decimal
+  let width = if size mod 4 = 0 then size / 4 else size / 4 + 1 in
+  let padding = String.make (width - (String.length word_str)) '0' in
+  Printf.sprintf "0x%s%s" padding word_str
 
 let format_values (fmt : Format.formatter) (vals : (string * z3_expr) list) : unit =
   let max_str_length =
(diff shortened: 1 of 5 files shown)


wp/lib/bap_wp/src/constraint.ml
```diff
@@ -49,6 +50,13 @@ let get_model_exn (solver : Solver.solver) : Model.model =
 let goal_to_string (g : goal) : string =
   Format.sprintf "%s: %s%!" g.goal_name (Expr.to_string (Expr.simplify g.goal_val None))
 
+let expr_to_hex (exp : z3_expr) : string =
+  let decimal = exp |> BV.numeral_to_string |> Int.of_string in
```
codyroux, line 54: Uh oh. Can this ever overflow? Isn't this 63 bits?

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -49,6 +50,13 @@ let get_model_exn (solver : Solver.solver) : Model.model =
...
+  let size = exp |> Expr.get_sort |> BV.get_size |> Int.to_float in
+  (* Calculate how many 0s to pad the output with. *)
```
codyroux, line 56: I guess this works for small numbers, but I don't understand why you don't use integer division and remainder... Like say `if size mod 4 = 0 then size/4 else size/4 + 1`. IDK

Floats make me nervous. Not really crucial though, since we're using this for printing.

wp/lib/bap_wp/src/constraint.mli
```diff
@@ -62,6 +62,9 @@ val goal_to_string : goal -> string
 (** Creates a string representation of a path. *)
 val path_to_string : path -> string
 
+(** Creates a string representation of a Z3 numeral in hexadecimal. *)
```
codyroux, line 65: Maybe add that this is in BAP notation?

wp/lib/bap_wp/src/constraint.ml
fortunac: Uh oh. It might be able to overflow.

wp/lib/bap_wp/tests/unit/test_constraint.ml
```diff
@@ -113,9 +113,18 @@ let test_substitute_order (test_ctx : test_ctxt) : unit =
   let goals = Constr.get_refuted_goals pre solver ctx in
   assert_refuted_goals test_ctx post_expr goals
 
+let test_expr_to_hex (test_ctx : test_ctxt) : unit =
+  let ctx = Env.mk_ctx () in
+  let expr = BV.mk_numeral ctx "1073741824" 64 in
+  let result = Constr.expr_to_hex expr in
+  let expected = "0x0000000040000000" in
```
codyroux, line 120: Any chance we can get an `0xffffffffffffffff` example?

codyroux:
Looks reasonable, except for the potential overflow, and that float stuff which isn't crucial to fix.

wp/lib/bap_wp/src/constraint.ml
fortunac: That honestly didn't come to mind when I wrote it. Having floats does make me nervous as well.

wp/lib/bap_wp/tests/unit/test_constraint.ml
fortunac: Added that test and it does overflow. Going to fix that now.

wp/lib/bap_wp/src/constraint.ml
codyroux: I just wrote that off the cuff though, you might want to test it...

fortunac: Yep, tested it and it works!

codyroux: 👍

PR #201, opened 2020-07-13
Phil/slim save

philzook58:
In the save-project plugin, removes attributes of terms as per Ivan's suggestion. This significantly reduces the size of the project files which in turn reduces the loading time of these project files. Down from ~600mb to ~45mb on a `tar` example and ~40s loading to ~5s. 

These are the attributes that are removed:

- alloc-size
- c.data
- c.proto
- c.type
- entry-point
- format
- insn
- malloc
- nonnull
- noreturn
- pure
- returns-twice
- warn-unused

The `address` attribute was kept for` --wp-print-paths`

### DIFF
--- a/wp/plugin/save_project/save_project.ml
+++ b/wp/plugin/save_project/save_project.ml
@@ -28,7 +28,10 @@ let main nm proj : unit =
   in
   let clear_mapper = object 
     inherit Term.mapper as super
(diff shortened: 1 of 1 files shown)


fortunac:
Looks good to me! If we need more attributes, we can add them back as needed. Please run `ocp-indent`, and I think it will be ready for a merge unless other people want to take a look.

philzook58:
Should be good now with respect to ocp-indent

PR #203, opened 2020-07-14
Functional loop unrolling depth

codyroux:
Instead of using a mutable hashtable, we now pass around a functional map, _within_ the `loop_handler` thunk contained in the `env`.

This should reduce the usual risks associated with mutability, e.g. if we re-use an existing `env` for a second analysis, as we are prone to doing.

Tests pass and I don't see any significant performance regression, but someone might want to double check that.


fortunac:
Makes sense to me!

Issue #186, opened 2020-06-30
Visiting empty subroutines


fortunac:
When visiting a subroutine, we assume that there is at least one block.

https://github.com/draperlaboratory/cbat_tools/blob/4652767e6385ad77295a9fb83b2ff69bcd2307c6/wp/lib/bap_wp/src/precondition.ml#L863

However, we've seen in our examples, some subroutines that do not have any blocks in them. In this case, we should just return the postcondition as the precondition.

fortunac:
We should also have a spec for these empty subroutines, which should be treated as a nop. The postcondition should be returned as a precondition.

fortunac:
Closed with #206.

PR #206, opened 2020-07-16
handling for empty subroutines

DieracDelta:
This PR addresses issue #186 by adding in an empty function spec. 

This passes all but the switch statement test (which is benign) in our test suite. When run on the grammatech examples, this turns 19 out of 23 SATs to UNSATs (the remainder take over 15 minutes on my computer and thus timeout).

### DIFF
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -513,6 +513,14 @@ let spec_verifier_nondet (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
   else
     None
 
+let spec_empty (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
+  if (Seq.is_empty @@ Term.enum blk_t sub) then
+    Some {
+      spec_name = "spec_empty";
+      spec = Summary (fun env post _tid -> post, env)
+    }
+  else None
+
 let spec_arg_terms (sub : Sub.t) (_ : Arch.t) : Env.fun_spec option =
   let args = Term.enum arg_t sub in
   if not (Seq.is_empty args) then
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -872,12 +872,14 @@ let filter (env : Env.t) (calls : string list) (cfg : Graphs.Ir.t) : Graphs.Ir.t
 
 let visit_sub (env : Env.t) (post : Constr.t) (sub : Sub.t) : Constr.t * Env.t =
   debug "Visiting sub:\n%s%!" (Sub.to_string sub);
-  let cfg = sub |> Sub.to_cfg |> filter env ["exit"] in
-  let start = Term.first blk_t sub
-              |> Option.value_exn ?here:None ?error:None ?message:None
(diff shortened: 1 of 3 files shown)


fortunac:
What happens if you try to inline those empty subroutines? Do we still get failures?

fortunac:
We're currently getting an error when we try to run WP directly on an empty subroutine. I think the correct behavior should be to return the postcondition as the precondition in this case. Can you look at
https://github.com/draperlaboratory/cbat_tools/blob/6557c633d0f80fe3c6c5b6c5d177f6b7ae7e3d66/wp/lib/bap_wp/src/precondition.ml#L865
and handle this? Thanks!

DieracDelta:
This makes sense. I put in a zero check (maybe I should be checking if the function spec is empty instead?) to `visit_sub` and reverted the spec precedence. Tests pass, and both inlining and analyzing an empty function no longer error.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -864,12 +872,14 @@ let filter (env : Env.t) (calls : string list) (cfg : Graphs.Ir.t) : Graphs.Ir.t
 
 let visit_sub (env : Env.t) (post : Constr.t) (sub : Sub.t) : Constr.t * Env.t =
   debug "Visiting sub:\n%s%!" (Sub.to_string sub);
-  let cfg = sub |> Sub.to_cfg |> filter env ["exit"] in
-  let start = Term.first blk_t sub
-              |> Option.value_exn ?here:None ?error:None ?message:None
-              |> Graphs.Ir.Node.create in
-  let pre, env' = visit_graph ~start:start env post cfg in
-  (pre, Env.add_precond env' (Term.tid sub) pre)
+  if (Seq.is_empty @@ Term.enum blk_t sub) then (post, env)
```
fortunac, line 875: I think we can add the postcondition to the environment over here. What do you think?

DieracDelta: Oops! Should be fixed now.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -864,12 +872,15 @@ let filter (env : Env.t) (calls : string list) (cfg : Graphs.Ir.t) : Graphs.Ir.t
...
 let visit_sub (env : Env.t) (post : Constr.t) (sub : Sub.t) : Constr.t * Env.t =
   debug "Visiting sub:\n%s%!" (Sub.to_string sub);
-  let cfg = sub |> Sub.to_cfg |> filter env ["exit"] in
-  let start = Term.first blk_t sub
-              |> Option.value_exn ?here:None ?error:None ?message:None
-              |> Graphs.Ir.Node.create in
-  let pre, env' = visit_graph ~start:start env post cfg in
-  (pre, Env.add_precond env' (Term.tid sub) pre)
+  let pre, env' =
+    if (Seq.is_empty @@ Term.enum blk_t sub) then (post, env)
```
codyroux, line 876: Do we want to print a warning here?

codyroux:
Looks good to me, merge at will.

wp/lib/bap_wp/src/precondition.ml
DieracDelta: I've added in a warning!

PR #208, opened 2020-07-17
Fortunac/print flags

### LINKED ISSUE #202: Rework print flags
Currently you need both --wp-print-paths and --wp-show-refuted-goals to get paths to print. This is confusing.

fortunac:
Fixes #202. Consolidated the different printing options into a single flag: `--wp-show=[bir|refuted-goals|paths|precond-internal|precond-smtlib]`. Passing in the option(s) to the flag will print to STDOUT the respective item(s).

- I did prepend `precond-` to `internal` and `smtlib` to specify that it is printing out the precondition since we no longer have the `print-constr` flag.
- If we pass in `paths`, we are assuming we are also printing out the refuted goals, because each path is for a refuted goal.
- I'm currently not too big on the magic strings we have for this flag, but that is a fix I want to leave for when we have a module for handling all of our flags.

### DIFF
--- a/wp/README.md
+++ b/wp/README.md
@@ -364,18 +364,6 @@ The various options are:
   follow the same execution trace. In the case the analysis returns UNSAT or
   UNKNOWN, no script will be outputted.
 
-- `--wp-print-refuted-goals[true|false]`. If set, in the case WP results in SAT,
-  prints a list of goals that have been refuted in the model. The list will show
-  the tagged name of the goal, the concrete values of the goal, and the Z3
-  expression representing the goal. For example, a refuted goal of
-  `(= RAX_orig RAX_mod)` can have concrete values of
-  `(= 0x0000000000000000 0x0000000000000001)`.
-
-- `--wp-print-path=[true|false]`. If present, will print out the path to a refuted
-  goal and the register values at each jump in the path. It also contains information
-  about whether a jump has been taken and the address of the jump if found. The path
-  will only be printed if refuted goals are printed with `--wp-print-refuted-goals`.
-
 - `--wp-use-fun-input-regs=[true|false].` If present, at a function call site, uses
   all possible input registers as arguments to a function symbol generated for
   an output register that represents the result of the function call. If set to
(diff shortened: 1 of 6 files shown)


wp/README.md
```diff
@@ -392,11 +380,21 @@ The various options are:
   original binary, then that same address is also non-null in the modified binary.
   Defaults to false.
 
-- `--wp-print-constr=[internal|smtlib] or both/neither`. If set, the preconditions
-  and Z3's SMT-LIB 2 are both printed. One or both outputs can be explicitly
-  called with the respective names `internal` and `smtlib`, which will print only
-  what is stated. Both can also be called like `--wp-print-constr=internal,smtlib`.
-  If the flag is not called, it defaults to printing neither.
+- `--wp-show=[bir|refuted-goals|paths|precond-internal|precond-smtlib]`. A list
```
codyroux, line 383: ❤️

wp/plugin/wp.ml
```diff
@@ -190,8 +188,8 @@ let analyze_proj (ctx : Z3.context) (var_gen : Env.var_gen) (proj : project)
   let pre, env = Pre.visit_sub env post main_sub in
   let pre = Constr.mk_clause [Z3_utils.mk_smtlib2_single env flags.pre_cond] [pre] in
   let pre = Constr.mk_clause hyps [pre] in
-  (* Print statement for constraint-style prover output: *)
-  Format.printf "\nSub:\n%s\nPre:\n%!" (Sub.to_string main_sub);
+  if List.mem flags.show "bir" ~equal:String.equal then
```
codyroux, line 191: I assume you'll have something better than string equality once this is all wrapped up in a module?

codyroux:
Looks like the user can enter whatever they want into the flag, and nothing happens if they make a typo, or enter a bunch of junk. Not strictly necessary, but we might want to filter by the list of accepted flags.

wp/plugin/wp.ml
fortunac: I'm thinking of something along the lines of:
```
type show =
  | Bir
  | RefutedGoals
  | Paths
  | PrecondInternal
  | PrecondSmtlib
```
that we can match on and possibly have it parsed into this type from the command line via http://binaryanalysisplatform.github.io/bap/api/odoc/bap/Bap/Std/Self/Config/index.html#val-converter

fortunac:
> 
> 
> Looks like the user can enter whatever they want into the flag, and nothing happens if they make a typo, or enter a bunch of junk. Not strictly necessary, but we might want to filter by the list of accepted flags.

I added a check for this!

Issue #26, opened 2019-06-19
Allow for "chaos" interpretation of expressions


codyroux:
Currently, we handle all operations in bit-precise manners, including shifts, adds, memory reads, etc. It might be a good idea, for specifications like "check that registers were correctly saved" to allow leaving some of these operations undefined, replacing them with uninterpreted functions.

This probably can be implemented by simply generalizing the hook mechanism for expressions: https://github.com/draperlaboratory/cbat_tools/blob/7d58dc86de8d6d3edcf72b79735b05cc207d623a/wp/lib/bap_wp/src/environment.mli#L72, say by adding an additional `Interp` cond_type.

codyroux:
We tried this, in PR #209 but did not observe a significant speedup.

PR #209, opened 2020-07-17
Adding a flag and utilites to convert BIR binary operations to uninterpreted Z3 functions, as mentioned in #26

gltrost:
A new type `and op_interp = Bap.Std.binop ->  Constr.z3_expr -> Constr.z3_expr -> Constr.z3_expr option` is added to `Environment.ml `, and is included in `Env.t`. 

In `Precondition.ml`, the functions `op_interp`, `op-uninterp`, `create_interp` and `op_interp_default` are created to allow for converting BIR binops to uninterpreted Z3 functions. 

A flag is created in `wp.ml` for allowing users to specify which operations they want uninterpreted, invoked with `--op-uninterp=[plus|sub|et. seq.]`. 

Known Issues and Things Worth Discussing:

1. It's not clear if there is a signifigant enough speed-up to include this feature. This still needs to be tested.
2. There may be a simpler and better way to create uninterpreted functions than how they are created with `funcDecl_to_z3`, `opaq_name`, and `create_interp`. 
3. `Bap.Std.Tid` was not used in creating uninterpreted functions, but may be helpful. 
4. Stylistically, the naming-conventions could be better. Betweeen `op_interp`, `op-uninterp` and `create_interp`, the naming convention could be clearer. 
5. The function `op_interp_default` may not be necesary.


wp/README.md
```diff
@@ -417,6 +417,13 @@ The various options are:
   be denoted in bytes. By default, the size of the stack is 0x800000, which is
   8Mbs.
 
+- `--op-uninterp=[plus|sub|mul|udiv|et.seq.] or some comma-delimited combination`.
```
fortunac, line 420: I'm not sure if we should have a default when calling this flag with no arguments. I'm expecting that no arguments means that we're not making anything uninterpreted. 

Can you also add a quick explanation and/or example explaining what uninterpreted means?

wp/lib/bap_wp/src/environment.ml
```diff
@@ -230,6 +233,7 @@ let init_loop_unfold (num_unroll : int) : loop_handler =
       in
       unroll
   }
+(*ADDED*)
```
fortunac, line 236: Was there supposed to be something to go along with this comment?

wp/lib/bap_wp/src/environment.ml
```diff
@@ -367,6 +373,10 @@ let get_sub_handler (env : t) (tid : Tid.t) : fun_spec_type option =
 let get_jmp_handler (env : t) : jmp_spec =
   env.jmp_handler
 
+(*ADDED*)
```
fortunac, line 376: Sorry for this tiny nitpick, but can we remove this comment?

wp/lib/bap_wp/src/environment.mli
```diff
@@ -211,6 +216,9 @@ val get_sub_handler : t -> Bap.Std.Tid.t -> fun_spec_type option
     jumps in a BIR program. *)
 val get_jmp_handler : t -> jmp_spec
 
+(** NOTE_TO_SELF: FILL ME IN *)
```
fortunac, line 219: Can we update the comment here?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -63,6 +60,30 @@ let binop (ctx : Z3.context) (b : binop) : Constr.z3_expr -> Constr.z3_expr -> C
...
+    let bop = match b with
+      | LSHIFT -> BV.mk_shl
+      | RSHIFT -> BV.mk_lshr
+      | ARSHIFT -> BV.mk_ashr
+      | _ -> assert false in
+    (* In x86 decoding, it is possible to scale the address with a 2-bitwidth shift
+       of 0, 1, 2, or 3. However, Z3 requires requires the operands of a bit shift
+       to be of the same bitwidth. Here, we pad the operand with the smaller
+       bitwidth to match the bitwidth of the other operand. *)
+    let get_size v = v |> Expr.get_sort |> BV.get_size in
```
fortunac, line 74: Thank you for moving this here! 🎉

wp/plugin/wp.ml
```diff
@@ -170,7 +171,9 @@ let analyze_proj (ctx : Z3.context) (var_gen : Env.var_gen) (proj : project)
   let specs = fun_specs flags to_inline in
   let exp_conds = exp_conds_mod flags in
   let stack_range = set_stack flags in
-  let env = Pre.mk_env ctx var_gen ~subs ~arch ~specs
+  let op_interp = (if (flags.op_interp = []) then Pre.op_interp_default
```
fortunac, line 174: Can we move this computation for the value of op_interp into a separate function so we can reuse it in both the single and comparative cases without having to maintain both of them?

wp/wp.ml
```diff
@@ -0,0 +1,515 @@
+(***************************************************************************)
```
fortunac, line 1: Was this file accidentally checked in?

fortunac:
Good job! I had a few nitpicks, but otherwise this looks great. 

Regarding some of the points you brought up:

> Bap.Std.Tid was not used in creating uninterpreted functions, but may be helpful.

I think we probably won't need the tids for the uninterpreted functions. The tids of the binops across the two binaries will differ, so I am having a hard time trying to imagine what we would use them for.

> Stylistically, the naming-conventions could be better. Betweeen op_interp, op-uninterp and create_interp, the naming convention could be clearer.

I got a bit confused when I saw the flag was named `op-uninterp`, but its value was stored in `op_interp`. Is that array storing operations you want to remain uninterpreted? Or is it operations you want interpreted? Or maybe I'm reading this completely wrong.

wp/lib/bap_wp/src/environment.ml
gltrost: Sorry. I thoght I had all the comments and (*ADDED*)'s removed. I'll get rid of all of those.

wp/plugin/wp.ml
gltrost: Great idea.

wp/lib/bap_wp/src/environment.ml
fortunac: 👍

wp/README.md
gltrost: That makes sense. I'll make the default the empty-list.
The example/explaination idea is good, since the concept is not intuitive to most.

wp/plugin/wp.ml
gltrost: Actually, I'm going to instead remove the computation for the _non_-comparative case, since the only time this is used is in comparisons.

fortunac: Sounds good! If it's only used for comparisons, I think we should have a note about that in the README and docs for the flag.

gltrost:
>I got a bit confused when I saw the flag was named op-uninterp, but its value was stored in op_interp. Is that array storing operations you want to remain uninterpreted? Or is it operations you want interpreted? Or maybe I'm reading this completely wrong.

The array stores what will be uninterpreted. I'll change this to `op_uninterp`.

wp/wp.ml
gltrost: Yes. I _think_ this is a line ending issue.

gltrost:
From the recommendation of @codyroux , I've just also added a test file `.../sample_binaries/op_uninterp_speedup`. However I didn't add any new test-cases to `.../plugin/tests/test_wp.ml` since the test file I added is mostly made to demonstrate performance differences when the `--wp-op-uninterpret` flag is on. 
If advised, I can add these new cases to `test_wp.ml` or move this new file to another location.

PR #210, opened 2020-07-20
Path Printing Bug

philzook58:
During small refactoring in previous commit, did not turn `List.forall` into `List.exists` correctly. This caused a bug in path printing.


fortunac:
Looks good! I'm going to merge this right away since it's a small change.

PR #212, opened 2020-07-22
Add set_to_eqs to compare_smtlib

fortunac:
Adds the hypothesis that registers are equal to each other at the beginning of a subroutine's execution.


codyroux:
Looks good, we might want to have a test in addition. Good job Chloe as usual, you're indispensable.

PR #214, opened 2020-07-23
Remove spurious 'dry run' in wp.ml

codyroux:
This should make the loop counter for unfolding take the correct values.\

I suggest @fortunac as a reviewer, since only she and I might remember why that code was there in the first place. Suspiciously, all the tests still pass...


fortunac:
We originally had that code to add all the z3 variables to the environment so a user can create a custom postconditions. Later on, we added the `init_vars` to the analysis and that function also adds the variables to the environment (along with their init counterparts), so it should be safe to delete the initial dry run.

fortunac:
I think we can delete the `Env.clear_consts` api because its whole purpose was to delete the constants generated during the dry run. It isn't used anywhere else. I don't think we'll need it anymore. What do you think?

codyroux:
I agree. We should remove `clear_consts`, though I can imagine situations where it might not be crazy to have, so maybe keep it around until we get to refactoring `Env.t`.

As far as I'm concerned, we don't need to deep dive on this any more, and can just merge.

Issue #188, opened 2020-07-01
Default callback for indirect calls


fortunac:
Currently, we have callbacks for direct function calls known as `fun_specs` https://github.com/draperlaboratory/cbat_tools/blob/d219cbed6609329ce5e76831e7c7103e56e5f602/wp/lib/bap_wp/src/environment.ml#L64
which takes in an `env`, `postcondition`, and `tid` of the target.

We should add a spec for indirect calls to the environment as well. Indirect calls do not have a `tid` for the target, but rather have an `exp`. Right now, we only need a default indirect function spec modeled similarly to the direct `spec_default`. 

https://github.com/draperlaboratory/cbat_tools/blob/d219cbed6609329ce5e76831e7c7103e56e5f602/wp/lib/bap_wp/src/precondition.ml#L620

At a function call site, WP looks up the function spec handler inside of `lookup_sub`
https://github.com/draperlaboratory/cbat_tools/blob/d219cbed6609329ce5e76831e7c7103e56e5f602/wp/lib/bap_wp/src/precondition.ml#L89
where it queries the environment for the fun_spec to be used. And calls it to determine the precondition. We should look up our indirect function specs here.

fortunac:
Closed with #219.

PR #219, opened 2020-07-28
Added indirect call spec

DieracDelta:
This PR addresses issue #188 . I'm not 100% sure what the default spec ought to be for indirect calls, but I've left it as incrementing the stack stack pointer. This doesn't seem to break any of the grammatech examples, and fixes some of the ones where a return address and return TID are provided. I think that generally:

- if there is a return tid accompanying the call, then there is a return address pushed onto the stack, and RSP should be incremented. Perhaps this is too strong of an assumption.
- if there is not a return tid accompanying the call, then the value of RSP will not be calculated correctly (the spec is wrong), because no return address is pushed onto the stack and thus RSP should not be incremented. However, in both programs, RSP is off by the same amount, resulting in preservation of the sats and unsats.

If the first assumption is not too strong, perhaps it makes sense add to the arguments to lookup_sub to distinguish whether or not the call is with a return. Alternatively, maybe the first assumption is too strong, and an empty spec should be applied in both cases. I'm not sure what makes the most sense here.


wp/lib/bap_wp/src/environment.ml
```diff
@@ -78,6 +79,8 @@ and fun_spec = {
   spec : fun_spec_type
 }
 
+and indirect_spec = t -> Constr.t -> Constr.t * t
```
fortunac, line 82: I think it might be a good idea to include the `Exp.t` as an argument in case we want to use that information in the spec in the future.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -95,9 +95,11 @@ let lookup_sub (label : Label.t) (post : Constr.t) (env : Env.t) : Constr.t * En
       | Some Inline -> !inline_func post env tid
       | None -> post, env
     end
-  (* TODO: Evaluate the expression for the indirect jump and
-   * figure out how to handle this case *)
-  | Indirect _ -> post, env
+  | Indirect exp ->
+    begin
+      warning "Encountered indirect call";
```
fortunac, line 100: Can you print out the exp in the warning?

fortunac:
I think we should add a simple test case to our test suite that has an indirect call in it.

DieracDelta:
The model example I was using to test this was `preorder` from grammatech. The invocation I used was (from the grammatech cbat docker container) was:

```bash
                bap /bin/echo \
                --pass=wp \
                --wp-compare=true \
                --wp-file1=orig.bpj \
                --wp-file2=mod.bpj \
                --wp-function=$OUTPUT \
                --wp-num-unroll=0 \
                --wp-inline=false \
                --no-byteweight \
                --wp-use-fun-input-regs=false \
                --wp-mem-offset=true \
                --wp-compare-post-reg-values=R12,R13,R14,R15,RBX,RSP,RBP,RAX \
                --wp-show=bir,refuted-goals,paths \
```
with `preorder` as `names.txt`.

DieracDelta:
The list of SAT symbols I pulled from grammatech are:
```
hash_find_entry
hash_do_for_each
seek_error_details
more_options
xdup2
_obstack_begin_worker
run_override_list
_wsplt_alloc_die
exclusion_tag_warning.part.5
preorder
read_warn_details
postorder
wordsplit_dump_nodes
decx
call_arg_warn
hash_lookup
undo_last_backup
file_removed_diag
xheader_read
safe_hasher
get_gnu_dumpdir
exclude_fnmatch
symlink_error
check_exclusion_tags
init_volume_number
call_arg_error
chmod_error_details
decode_time
assign_time_option
write_long_link
call_arg_fatal
run_decompress_program
read_error_details
set_use_compress_program_option
seek_warn_details
read_timespec
unknown_file_error
argp_version_parser
merge_sort_sll
check_links
_wsplt_nomem
read_and_process.constprop.6
read_fatal_details
info_attach_exclist
get_date_or_file.isra.6
frame_dummy
sparse_fixup_header
write_error_details
link_error
_init
_start
```
A combination of `mem-offset=false` and this PR turns about half of these examples to UNSAT. I took these stats about a week ago though, and I'm not sure where the results went. I'll need to rerun the examples to get exact numbers.

fortunac:
Looks good! Just those little nitpicks and adding a test, and I think this should be good to go.

DieracDelta:
Okay, I've added:
- an more detailed warning that prints the expression target of the indirect call
- `Exp.t` to the spec
- a simple test case

fortunac:
I tried running wp on one of the binaries in the test with the custom postcondition `(assert (= RSP (bvadd init_RSP #x0000000000000008)))` expecting UNSAT, but I got SAT instead. It looks like we are interpreting returns as indirect calls because returns are lifted into BIL as

```
#1 = mem[RSP, el]:u64
RSP := RSP + 8
call #1 with noreturn
```

This causes RSP to be incremented by another 8 even though we are already incrementing it in the 2nd line. I'm not sure what the best way to solve this is. We should have a discussion regarding this.

DieracDelta:
Upon further discussion and per Cody and Chloe's suggestions, the `increment_stack_pointer` spec should only be applied if there is an indirect call with a return. This consistently pushes the return address onto the stack in all cases in the grammatech example. If there is not a return included, then an empty spec should be applied. Implementing this involves four changes:

- First the logic from `visit_jmp` in the case of a `Call` will be moved to `lookup_sub`
- Second `lookup_sub` will be renamed to `visit_call`, since its job has become broader.
- Changing the included test case to assert that the RSP at the end of the function matches the RSP at the beginning of the function.
- Adding a second test case with no return.

DieracDelta:
I've pushed the four changes outlined above. It's worth noting that I added a boolean to the default spec that carries whether or not the indirect call has a return on it. I'm not sure if this is the right place for this. I could alternatively move the stack increment into the `visit_call` invocation after the indirect call spec has been applied. The question here is: "should the return part of the indirect call spec?" Incrementing RSP is done by the indirect call, but whether or not the indirect call has a return is how we tell, and this isn't part of the indirect call.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -321,6 +308,37 @@ let increment_stack_ptr (post : Constr.t) (env : Env.t) : Constr.t * Env.t =
...
+  | Some (Summary compute_func) -> compute_func env post tid
+  | Some Inline -> !inline_func post env tid
+  | None -> post, env
+
+let visit_call (call: Bap.Std.Call.t) (post : Constr.t) (env : Env.t) : Constr.t * Env.t =
+  let target = Call.target call in
+  let return = Call.return call in
+  match target, return with
+  | Direct _, Some (Indirect _)
+  | Indirect _, Some (Indirect _) -> post, env
```
fortunac, line 329: Can we add a warning message that we have an indirect return here?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -321,6 +308,37 @@ let increment_stack_ptr (post : Constr.t) (env : Env.t) : Constr.t * Env.t =
...
+  | Indirect t_exp, None ->
+    Env.get_indirect_handler env t_exp env post t_exp false
+  | Direct t_tid, None ->
+    lookup_sub_handler t_tid env post
+  | Direct t_tid, Some (Direct r_tid) ->
+    let ret_pre = lookup_precond r_tid env post in
+    lookup_sub_handler t_tid env ret_pre
+  | Indirect t_exp, Some (Direct r_tid) ->
+    let ret_pre = lookup_precond r_tid env post in
+    Env.get_indirect_handler env t_exp env ret_pre t_exp true
```
fortunac, line 339: Can we add a log that we are visiting an indirect call for this case and the indirect case above?

fortunac:
The changes look good to me! Just add the few logs, and it should be good to go.

> 
> 
> I've pushed the four changes outlined above. It's worth noting that I added a boolean to the default spec that carries whether or not the indirect call has a return on it. I'm not sure if this is the right place for this. I could alternatively move the stack increment into the `visit_call` invocation after the indirect call spec has been applied. The question here is: "should the return part of the indirect call spec?" Incrementing RSP is done by the indirect call, but whether or not the indirect call has a return is how we tell whether or not RSP should be incremented. This isn't part of the indirect call; rather it's part of the indirect call context.

I think in the future, we might want to separate this out into two different specs. One spec that is just a nop and another spec that increments the stack pointer. Then, `Env.get_indirect_handler` can look up the right spec based on whether there is a return or not. I'm fine with leaving this for future work.

DieracDelta:
Added prints!

Issue #218, opened 2020-07-28
Integrate the collator interface with WP


fortunac:
This is related to #144. In order to simplify WP invocation, the following changes are needed:

1. Update the WP plugin to become a BAP command.
```
Bap_main.Extension.Command.declare grammar name command
```
The grammar is the arguments we can accept. The name is just "wp", and command will be our main function.
We can continue reading our bpj files at this point.

2. Update the command to load the binaries directly with
```
let input = Project.Input.file ~loader ~filename in
Project.create ~package input
```
We also need to manually run the api pass on both projects with
```
match Project.find_pass "api" with
| Some pass -> Project.Pass.run_exn pass proj
```
3. Add caches for loading a project. Look at [here](https://github.com/draperlaboratory/CBAT-internal/tree/master/code/BAP/bap_examples_2.1.0/06_cmd_with_caching) on how to do this.
4. We need to define a collator (possibly in a different file)
```
 Project.Collator.register name ~package ~prepare ~collate ~summary ~desc;
```
which will need three main functions. The collator is basically a giant fold on our projects. `prepare` is run on the original binary. `collate` takes in the result of prepare and is run on the modified binaries. When collate is completed, `summary` is run. We can place our output functions there.
5. Back in the original extension, we have to apply the collator with
```
match Project.Collator.find ~package name with
| Some collator -> Project.Collator.apply collator projs
```

PR #220, opened 2020-07-28
Fortunac/cmd arguments

fortunac:
Integrates parts 1-3 in #218. This is a very big update that transforms our wp plugin to a bap command. All the `run_wp` scripts in the internal repo will break. 

### Interface changes
- WP's invocation is now `bap wp --func=foo [options] file1 [file2]`
- All the flags are no longer prepended with `wp-`
- Some flag names have been changed to prevent ambiguity:
  - `check-calls` => `compare-func-calls`
  - `gdb-filename` => `gdb-output`
- Removed some flags that are no longer necessary:
  - `compare`, `file1`, and `file2` in favor of WP determining whether to perform a single or comparative analysis depending on the number of input files.
- To view the help page, the command is `bap wp --help`.

### Other changes/notes
- Updated the README with the new flags.
- This PR still reads from our `.bpj` files. The upcoming PR will remove the `save_project` plugin, so we will no longer need this.
- The majority of the changes are located in `plugin/wp.ml`.  It looks like a lot of changes, but it is all the same functions, but placed in separate modules.


codyroux:
I'm a little worried about "guessing" behavior in options, generally: if the user forgets to put `file2`, but asks for `compare-func-calls`, what happens?

fortunac:
> 
> 
> I'm a little worried about "guessing" behavior in options, generally: if the user forgets to put `file2`, but asks for `compare-func-calls`, what happens?

That property won't be added because we only check this flag in the comparative case, and wp continues running. I can add a check for the flags the assume we have two binaries. It will let the user know that the flag is meant for the comparative case and the user did not specify a second file, and then exit.

Issue #171, opened 2020-06-16
Add examples into tests


philzook58:
Some of @DieracDelta examples in the documentation are really great. It'd be nice if we could add them into our test suite

DieracDelta:
Closed with #228

PR #223, opened 2020-07-30
additional tests

DieracDelta:
Addresses #171. This brings in a bunch of tests that have tested the scalability of cbat. Note that they also introduce a dependency to run the test suite on `gcc-8` for the loop unrolling pragma. This may be installed with:

```
sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
sudo apt-get -y update
sudo apt-get -y install gcc-8 g++-8
```

The tests involved include:
- a hash function.
- a solver for the n-queens problem. Currently set to `n=4` for quick completion.
- a 4 by 4 sudoku solver.
- a comparison of of LSB calculations between two binaries. The first method uses debruijn sequences and the second method uses a for loop.


wp/resources/sample_binaries/debruijn/main_1.c
```diff
@@ -0,0 +1,107 @@
...
+                 25, 14, 19,  9, 13,  8,  7,  6
+        };
+
+        uint64_t debruijn64 = 0x03f79d71b4cb0a89;
+
+        return index64[((bb & -bb) * debruijn64) >> 58];
+}
+
+// calculate the index of the least significant bit (128 bit version)
+#define uint128_t __uint128_t
```
gltrost, line 73: Is is still worth having the code for `uint129_t` size testing? My recollection was that cbat couldn't handle this large of a test case.

wp/resources/sample_binaries/debruijn/main_1.c
```diff
@@ -0,0 +1,107 @@
+#include <assert.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdio.h>
+
+// taken from http://supertech.csail.mit.edu/papers/debruijn.pdf
```
gltrost, line 6: Would it be worth having a small sentence summarizing this pdf? The paper itself is somewhat involved for anyone trying to understand the problem.

DieracDelta:
Okay I've removed the `128` bit code because it's too much for now. I've also added in a readme for debruijn that provides a brief summary of how it works. Finally, I've removed the gcc-8 dependency (should now work with any version of gcc!).

wp/resources/sample_binaries/debruijn/main_2.c
```diff
@@ -0,0 +1,80 @@
+#include <assert.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdio.h>
+#pragma GCC push_options
+#pragma GCC optimize ("unroll-loops")
+
```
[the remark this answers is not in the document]
WithoutCaps: Update:
Now that i know what is the reason for all this it makes sense. 

Good Job, buddy!

DieracDelta:
This is not ready to be merged. There are two outstanding issues (thanks @gltrost for pointing these out!):
- nqueens has multiple solutions. All these solutions should be added as options so the test always passes.
- Removing the gcc-8 dependency results, and using gcc-5 with a different set of pragmas instead results in the 8 and 16 bit debruijn loops being completely unrolled. However the 32 and 64 bit loops do not get unrolled completely.

DieracDelta:
I've unrolled the 32 and 64 bit loops manually for debruijn. The issue now is that there is an implicit loop in the 64 bit version with debruijn sequences. WP does not finish in a reasonable amount of time when manually unrolling the loop 64 times. GCC-8 does not have this problem. Since we're encountering performance problems with the 32bit version, I'm going to delete the 64bit version for now.

fortunac:
I'm getting SAT for Debruijin 64-bit. Am I doing something wrong here?

DieracDelta:
Oops I forgot to upstream my changes (they should be pushed now). I'm removing 64bit debruijn for the time being so we don't have a dependency on gcc-8. Note that this isn't ready to be merged; it needs to be rebased on top of #228

DieracDelta:
Okay, I've rebased onto master! Nqueens has two solutions (which are now both included).

wp/resources/sample_binaries/debruijn/run_wp_16bit.sh
```diff
@@ -0,0 +1,19 @@
+# an example that compares the debruijn method of finding the LSB instead
```
fortunac, line 1: Can we add a quick comment about what the expected result (SAT vs UNSAT) is inside the run files so we don't have to look them up in the test suite when we are running them directly?

DieracDelta:
I've added comments on the expected output to each `run_wp*` script!

fortunac:
Looks good to me! I think `wp/resources/sample_binaries/hash_function/.run_wp.sh.swp` was accidentally checked in, but other than that, I think it's all good! Does @gltrost have any more input?

DieracDelta:
Oops. Should be unstaged now :+1:

Issue #225, opened 2020-08-05
multiple models for unit tests


DieracDelta:
Currently in our tests, we assume that there is one correct model when invoking `test_plugin` in `test_wp.ml`. We should add a way to provide multiple models. This will allow us to more precisely test the correctness of tests that have multiple solutions, such as nqueens and sudoku. 

To do this, a simple approach would be to include a list of models of type `((string * string) list) list` as `~expected_regs` to `test_plugin` and propagate this down to `check_model`, where we would perform several iterations over the output (instead of one).

An example parameter for `expected_regs` would be:

```
~expected_regs:[[("RSI", "0x0000000000000010"); ("RCX", "0x0000000000000007")]; [("RSI", "0x0000000000000010"); ("RCX", "0x0000000000000008")];]
```

This would allow us to check that the model either matches `RSI = 0x10 and RCX = 0x7` or `RSI = 0x10 and RCX = 0x8`.

fortunac:
Closed with #228.

PR #228, opened 2020-08-11
Allow list of countermodels for tests

DieracDelta:
This PR addresses issue #225 by allowing a list of countermodels to be passed into our test suite. If any one of those countermodels matches, the test will pass. If none of the models match, the mismatched registers in each model are printed. This will allow us to pass in all solutions to a sudoku puzzle or nqueens and thereby always pass even in the face of the varying solutions across machines. This should allow for #223 to be rebased on top of this commit.

### DIFF
--- a/wp/plugin/tests/test_wp.ml
+++ b/wp/plugin/tests/test_wp.ml
@@ -39,42 +39,78 @@ let check_z3_result (line : string) (expected : string) (ctxt : test_ctxt) : uni
   else
     ()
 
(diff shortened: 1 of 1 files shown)


fortunac:
Looks good! Do we have any examples that we can add to the test suite that have multiple models we can check for?

fortunac:
Oh wait, I didn't see the top comment. Looks good!

DieracDelta:
The examples that use this are in #223 (though, I've tested simpler with toy examples that are not included in this commit).

PR #229, opened 2020-08-13
Phil/smt compat

philzook58:
In experimenting with other solvers, I've found that z3 has different conventions with respect to `and` nodes and possibly other operations. Other solvers do not parse zero or single argument ands. In addition, other solvers do not support the `bvredor` bitvector instruction.

### DIFF
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -57,8 +57,8 @@ let binop (ctx : Z3.context) (b : binop) : Constr.z3_expr -> Constr.z3_expr -> C
   | AND -> BV.mk_and ctx
   | OR -> BV.mk_or ctx
   | XOR -> BV.mk_xor ctx
-  | EQ -> fun x y -> BV.mk_not ctx @@ BV.mk_redor ctx @@ BV.mk_xor ctx x y
-  | NEQ -> fun x y -> BV.mk_redor ctx @@ BV.mk_xor ctx x y
+  | EQ -> fun x y -> Bool.mk_ite ctx (Bool.mk_eq ctx x y) one zero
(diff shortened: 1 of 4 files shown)


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -57,8 +57,8 @@ let binop (ctx : Z3.context) (b : binop) : Constr.z3_expr -> Constr.z3_expr -> C
   | AND -> BV.mk_and ctx
   | OR -> BV.mk_or ctx
   | XOR -> BV.mk_xor ctx
-  | EQ -> fun x y -> BV.mk_not ctx @@ BV.mk_redor ctx @@ BV.mk_xor ctx x y
-  | NEQ -> fun x y -> BV.mk_redor ctx @@ BV.mk_xor ctx x y
+  | EQ -> fun x y -> Bool.mk_ite ctx (Bool.mk_eq ctx x y) one zero
```
fortunac, line 60: Are there any performance impacts, both speed and memory, on changing the `mk_redor` to a `mk_ite`?

philzook58: This is something I should have information on but don't. Anecdotally I did not notice much. I don't think this code gets executed that many times based on the number of `bvredor` I was finding in the smtlib2 output (~ 5) in my examples

fortunac: Do you know how to memory performance differs?

philzook58: If by memory performance you mean total memory usage, I have not noticed memory usage to be a limiting factor at any point except when were were holding onto too much information in the project file.

philzook58: I am not sure if this branch should be pulled in at the moment until these questions are resolved. I made a pull request for comments and visibility if anyone else wants to try using external smt solvers.

philzook58:
Subsumed by #244

Issue #216, opened 2020-07-27
Add pointer register value constraints


DieracDelta:
Following issues #213 and #190 and the discussion this morning, a necessary first step would be to, given a list of registers designated as pointers, restrict their values to only initialized memory. At first, this would mean (to the point of issue #213) preventing these pointers from going past RSP. Further down the line, we may wish to extend this to include other regions, and be conscious of memory accesses that are at an offset.

This includes two independent parts:

- Include a function summary for each function that includes which registers are pointers. We would want to store this in the knowledge base once it is ready.
- Figure out a way to, for a given function and set of pointer registers, add in a constraint on the value of these marked pointer registers. Some strawman implementation details mentioned in the meeting earlier today include:
    - Feeding in a set of registers to WP for the function to be analyzed. We may temporarily add in a flag to do this.
    - Restrict the value of those registers at the start of the function analysis using a z3 expression (rather similarly to adding these restrictions in as a precondition). For now, constraining the pointer value simply means "don't go past the current RSP." Since RSP only decreases during the function execution, this should (theoretically) prevent issue #213.

ccasin:
This looks right to me!  Of the two independent parts you mentioned, I think focusing on the second makes sense (and coming up with some minimal effort hack for the first part just so you can test the second part).  Presumably our view on function summaries/the knowledge base will change how we want to do the first part when Chloe finishes adapting to Ivan's improved interface soon.

DieracDelta:
The feature described by the second bullet point is now merged.

Issue #213, opened 2020-07-22
return address 


DieracDelta:
I've got [main_1](https://gist.github.com/98ea38abde6afa9f4b346db123e2ccc5), [main_2](https://gist.github.com/733c72e06888f8dc637cf6096ad2c294), an [invocation](https://gist.github.com/bc22ed148646846cd57717faeefa46b5), the main_1 [cfg](https://gist.github.com/cf91dfe7ff8d0a544035a7957e156a4f), and the main_2 [cfg](https://gist.github.com/4b7e8f3ac0ad6a7563533bff758fed4a). This comes back SAT with [this](https://gist.github.com/e0f48ba8f06042a69ca2f17284771a3c) countermodel and it looks like the return address being pushed onto the stack right before the function call (tid %00000026) is the culprit. I believe this is because CBAT chooses a value for RBX that is on the stack and includes part of this return address. mem[rbx] is then read into a temporary variable. And then that temporary variable is then moved into RAX, which is our refuted goal. I did confirm that the tid has a spec (_start has `spec_arg_terms`), and thus stack pointer is incremented. However, if one removes the offset label and instruction in `main_2` that cause the pushed return address to be different, the programs are unsat. Does this make sense?

This feels like something that can be corrected by adding in an assertion afterwards that the stack memory below RSP after a function is called is equal between programs. However, maybe this is this too strong of an assumption?

DieracDelta:
I'm pasting in the gist metadata. `main_1`'s code :
```
.global _start

.text

safe_hasher:

push %rbx
#callq *0x30(%rbx)
callq _start
add (%rbx),%rax
pop %rbx
retq

main:
    ret

_start:
    call    main
```
`main_2`'s code:
```
.global _start

.text

this_creates_an_offset:
push %rbx

safe_hasher:

push %rbx
#callq *0x30(%rbx)
callq _start
add (%rbx),%rax
pop %rbx
retq

main:
    ret

_start:
    call    main
```
`main_1`'s cfg:
```
digraph {
node[shape=box];

subgraph "cluster__start" {
  label="_start"
  
  "\%0000001c"[label="\%0000001c
0000002f: RSP := RSP - 8\l00000032: mem := mem with [RSP, el]:u64 <- 0x4000E5\l00000035: call @main with noreturn\l"]
}
"\%0000001c" -> "\%0000002b";

subgraph "cluster___libc_start_main" {
  label="__libc_start_main"
  
[...]

PR #230, opened 2020-08-17
mark pointer inputs with flag

DieracDelta:
This PR introduces a flag, `--wp-pointer-reg-list` and closes the second half of #216 . The purpose of this flag is to, at the start of a function, mark registers as pointers. These specially marked registers are then used to generate a precondition that restricts their value to anywhere that is not uninitialized stack. That is to say, these "pointer" register values are either above the initial value of `RSP` or below the bottom of the stack. A sample invocation is included with the example, [here](https://github.com/DieracDelta/cbat_tools/blob/jrestivo/precond_flag/wp/resources/sample_binaries/pointer_flag/run_wp_unsat.sh).

I've tested this on the grammatech examples from #213 , and the simple toy example included in this commit.
- The toy example: CBAT is unable to use differing return addresses to obtain differing return values with the flag (the inspiration for this PR), but comes back SAT without the flag.
- `is_dumpdir` remains SAT when `RDI` is marked as a pointer. This is because in execution, there are accesses at an offset, which result in `wp` choosing a particular `RSP` such that `RDI + $OFFSET == pushed return address`.
- `safe_hasher` and `mb_copy` suffer from the same issue as `is_dumpdir`, though they're more subtle.

Note that I've only included a test for the comparative case (and double checked that the single analysis case doesn't break things). It's not obvious to me how to obtain different results (SAT vs UNSAT) with and without this flag in a single binary analysis.

### DIFF
--- a/wp/lib/bap_wp/src/compare.mli
+++ b/wp/lib/bap_wp/src/compare.mli
@@ -104,15 +104,15 @@ val compare_subs_empty_post : comparator * comparator
     creation time. *)
 val compare_subs_sp : comparator * comparator
 
-(** Compare two subs by composition for restraints on a pointer.:
+(** Compare two subs by composition for custom preconditions and postconditions.
 
     Give two subroutines and environments, return a postcondition
     and hypothesis that, when passed to [compare_subs], will generate a
     precondition which is provable only if (modulo soundness bugs) the VCs generated
-    by the hooks provided by the environment are satisfied, given that the
-    architecture of the binary is x86_64. The hypothesis comparator generates
-    hypotheses that are the precondition; the post condition comparator
-    generates postconditions based on the input Constr.t. *)
+    by the hooks provided by the environment are satisfied, given that the.
+    The hypothesis comparator generates hypotheses that are the input
+    [pre_conds] argument. The post condition comparator generates postconditions
+    that are the input [post_conds] argument. *)
 val compare_subs_constraints
   :  pre_conds:Constr.t
   -> post_conds:Constr.t
--- a/wp/lib/bap_wp/src/compare.mli
+++ b/wp/lib/bap_wp/src/compare.mli
@@ -104,15 +104,10 @@ val compare_subs_empty_post : comparator * comparator
     creation time. *)
 val compare_subs_sp : comparator * comparator
 
-(** Compare two subs by composition for custom preconditions and postconditions.
-
-    Give two subroutines and environments, return a postcondition
-    and hypothesis that, when passed to [compare_subs], will generate a
-    precondition which is provable only if (modulo soundness bugs) the VCs generated
-    by the hooks provided by the environment are satisfied, given that the.
-    The hypothesis comparator generates hypotheses that are the input
-    [pre_conds] argument. The post condition comparator generates postconditions
-    that are the input [post_conds] argument. *)
+(** Given a constraint to use as the hypothesis (precondition) and a constraint
+    to use as the postcondition, returns a tuple of comparators. The first
+    comparator returns the input postcondition constraint when called. The
+    second comparator returns the input precondition constraint when called. *)
 val compare_subs_constraints
   :  pre_conds:Constr.t
   -> post_conds:Constr.t
--- a/wp/lib/bap_wp/src/compare.mli
+++ b/wp/lib/bap_wp/src/compare.mli
@@ -104,6 +104,19 @@ val compare_subs_empty_post : comparator * comparator
     creation time. *)
 val compare_subs_sp : comparator * comparator
 
+(** Compare two subs by composition for restraints on a pointer.:
+
+    Give two subroutines and environments, return a postcondition
+    and hypothesis that, when passed to [compare_subs], will generate a
+    precondition which is provable only if (modulo soundness bugs) the VCs generated
+    by the hooks provided by the environment are satisfied, given that the
+    architecture of the binary is x86_64. The hypothesis comparator generates a
+    constraint which states whichever TODO *)
+val compare_subs_constraints
+  :  pre_conds:Constr.t
+  -> post_conds:Constr.t
+  -> comparator * comparator
+
 (** Compare two subroutines by composition for conservation of function calls:
 
     Given two subroutines and environments, return a postcondition
--- a/wp/lib/bap_wp/src/compare.mli
+++ b/wp/lib/bap_wp/src/compare.mli
@@ -110,8 +110,9 @@ val compare_subs_sp : comparator * comparator
     and hypothesis that, when passed to [compare_subs], will generate a
     precondition which is provable only if (modulo soundness bugs) the VCs generated
     by the hooks provided by the environment are satisfied, given that the
-    architecture of the binary is x86_64. The hypothesis comparator generates a
-    constraint which states whichever TODO *)
+    architecture of the binary is x86_64. The hypothesis comparator generates
+    hypotheses that are the precondition; the post condition comparator
+    generates postconditions based on the input Constr.t. *)
 val compare_subs_constraints
   :  pre_conds:Constr.t
   -> post_conds:Constr.t
--- a/wp/lib/bap_wp/src/environment.mli
+++ b/wp/lib/bap_wp/src/environment.mli
@@ -251,6 +251,9 @@ val use_input_regs : t -> bool
     defined by the concrete range of the stack in the env. *)
 val in_stack : t -> Constr.z3_expr -> Constr.z3_expr
 
+(** [get_stack_bottom] retrieves the address of the bottom of the stack *)
+val get_stack_bottom : t -> int
+
 (** [in_data_section env addr] is the constraint [DATA_MIN <= addr < DATA_MAX] as
     defined by the concrete range of the data section in the env. *)
 val in_data_section : t -> Constr.z3_expr -> Constr.z3_expr
--- a/wp/lib/bap_wp/src/environment.mli
+++ b/wp/lib/bap_wp/src/environment.mli
@@ -231,6 +231,9 @@ val get_loop_handler :
 (** Obtains the architecture of the program. *)
 val get_arch : t -> Bap.Std.Arch.t
 
+(** Obtains the name of the program's stack pointer *)
+val get_rsp_name : t -> string
+
 (** Obtains a list of all the {!Constr.z3_expr}s that represents constants that
     were generated during analysis. *)
 val get_consts : t -> ExprSet.t
@@ -251,8 +254,13 @@ val use_input_regs : t -> bool
     defined by the concrete range of the stack in the env. *)
 val in_stack : t -> Constr.z3_expr -> Constr.z3_expr
 
-(** [get_stack_bottom] retrieves the address of the bottom of the stack *)
-val get_stack_bottom : t -> int
+(** [get_stack_end] retrieves the address of the largest address that the top
+     of the stack can take. It is assumed the stack grows downwards, from
+     high addresss to low address. The top and start of the stack is defined as
+     stack_base and is the largest address in the stack. The stack end
+     (retrieved by this function) is the smallest value that the stack pointer
+     can take. *)
+val get_stack_end: t -> int
 
 (** [in_data_section env addr] is the constraint [DATA_MIN <= addr < DATA_MAX] as
     defined by the concrete range of the data section in the env. *)
--- a/wp/lib/bap_wp/src/environment.mli
+++ b/wp/lib/bap_wp/src/environment.mli
@@ -232,7 +232,7 @@ val get_loop_handler :
 val get_arch : t -> Bap.Std.Arch.t
 
 (** Obtains the name of the program's stack pointer *)
-val get_rsp_name : t -> string
+val get_sp_name : t -> string
 
 (** Obtains a list of all the {!Constr.z3_expr}s that represents constants that
     were generated during analysis. *)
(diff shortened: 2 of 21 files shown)


DieracDelta:
I've taken @codyroux 's suggestions into account. Specific changes include:
- A rebase onto master (using the new cli)
- A single binary analysis test case for the flag
- Constraints are now constructed using the Z3 api only
- Comments are included in the `construct_pointer_constraint` docstring detailing the constraint being built

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -79,13 +79,13 @@ let build_str (tokens : string list) : string =
    [fmt] is used to add prefixes and suffixes to a variable name. For example,
    init_RDI_orig. *)
 let get_z3_name (map : Constr.z3_expr Var.Map.t) (name : string) (fmt : Var.t -> string)
-  : string option =
+  : (Z3.Expr.expr option) =
```
fortunac, line 82: Super small nit, can we use `Constr.z3_expr` here to be more consistent?

wp/lib/bap_wp/src/z3_utils.mli
```diff
@@ -32,7 +32,7 @@ val build_str : string list -> string
     [fmt] is used to add prefixes and suffixes to a variable name. For example,
     init_RDI_orig. *)
 val get_z3_name :
-  Constr.z3_expr Bap.Std.Var.Map.t -> string -> (Bap.Std.Var.t -> string) -> string option
+  Constr.z3_expr Bap.Std.Var.Map.t -> string -> (Bap.Std.Var.t -> string) -> (Z3.Expr.expr option)
```
fortunac, line 35: Here too.

wp/plugin/lib/parameters.ml
```diff
@@ -35,6 +36,8 @@ type t = {
   show : string list
 }
 
+(* TODO add in validate pointer_reg_list *)
```
fortunac, line 39: Are you planning on adding this function in this PR or later?

wp/resources/sample_binaries/pointer_flag/run_wp_unsat.sh
```diff
@@ -0,0 +1,24 @@
+# This is a test that tests that the wp-pointer-reg-list flag works as expected
+# specifically, it calls a function, then tries to use the differing return address that 
+# is thereby pushed onto the stack to cause a different return value between functions.
+# Should return SAT
+
+# Should return UNSAT
```
fortunac, line 6: Are we expecting a SAT or UNSAT for this test?

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -111,3 +111,58 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
   mk_smtlib2 ctx smt_post decl_syms
 
+let construct_pointer_constraint (l: string list) (env1 : Env.t)
+    (env2: Env.t option): Constr.t =
+  let stack_bottom = Env.get_stack_bottom env1 in
+  let ctx = Env.get_context env1 in
+  let arch = match Env.get_arch env1 |> Bap.Std.Arch.addr_size with
+    | `r32 -> 32
+    | `r64 -> 64 in
+  let rsp = "RSP" in
```
fortunac, line 121: Hardcoding the string "RSP" makes me a bit uncomfortable. Since we have the architecture, we can grab the name of the stack pointer:
```
let module Target = (val target_of_arch arch) in
let rsp = Var.name Target.CPU.sp in
...
```

wp/lib/bap_wp/src/environment.mli
```diff
@@ -251,6 +251,9 @@ val use_input_regs : t -> bool
     defined by the concrete range of the stack in the env. *)
 val in_stack : t -> Constr.z3_expr -> Constr.z3_expr
 
+(** [get_stack_bottom] retrieves the address of the bottom of the stack *)
```
fortunac, line 254: Can you add a bit of clarification that the stack is grows from top to bottom, and the stack bottom is not the same as its base?

wp/plugin/lib/parameters.ml
DieracDelta: Oops! Since this works on both single and comparative analysis, it's probably not necessary? I deleted the comment.

DieracDelta:
I've incorporated in @fortunac 's suggestions. A couple things I'm uncertain about:
- the explanation for `get_stack_bottom` is a bit confusing. I renamed the function to `get_stack_end` and changed comments. Hopefully this is less confusing, thought it's still a bit of a misnomer, since it's not exactly the end of the stack as much as the minimum value that the stack pointer can take/the largest size the stack can be.
- The placement of `construct_pointer_constraint` is in `z3_utils`. It's making ample use of the Z3 api, which is why I've placed it here.

fortunac (commented):
Looks good to me! Just a few minor details that I commented below. Cody might want to double check the z3 expression generated, but it looks good to me. Good job!

DieracDelta:
Added docs and fixed those makefiles. Good catch!

wp/lib/bap_wp/src/compare.mli
```diff
@@ -104,6 +104,20 @@ val compare_subs_empty_post : comparator * comparator
     creation time. *)
 val compare_subs_sp : comparator * comparator
 
+(** Compare two subs by composition for restraints on a pointer.:
```
codyroux, line 107: I don't think this is an appropriate description of this function: nothing in the body of the function mentions pointers in any way.

In general this function seems very generic, so I'm not sure why you're mentioning x86_64, or indeed why the description is what it is. It looks like this function just bottles up pre and post conditions. The comment should reflect that.

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -111,3 +111,59 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
 
+let construct_pointer_constraint (l: string list) (env1 : Env.t)
+    (env2: Env.t option): Constr.t =
+  let stack_end = Env.get_stack_end env1 in
+  let ctx = Env.get_context env1 in
+  let arch = match Env.get_arch env1 |> Bap.Std.Arch.addr_size with
+    | `r32 -> 32
+    | `r64 -> 64 in
+  let rsp = Env.get_rsp_name env1 in
+  let err_msg_input = "invalid register name in pointer register list" in
```
codyroux, line 122: Probably a good idea to have the name of the function in the error message, in my experience.

codyroux: Like `"construct_pointer_constraint: invalid register name in pointer register list"`

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -111,3 +111,59 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
+  let err_msg_rsp = "stack pointer not found" in
```
codyroux, line 123: And same here.

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -111,3 +111,59 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
+          (get_z3_name init_var_map_orig rsp Var.name) in
+      let rsp_mod = Option.value_exn ~message:("modified " ^ err_msg_rsp)
+          (get_z3_name init_var_map_mod rsp Var.name) in
+      (* Encode constraint that each register is not within stack*)
+      (fun acc reg ->
+         let reg_name_orig = Option.value_exn ~message:err_msg_input
+             (get_z3_name init_var_map_orig reg Var.name) in
+         let reg_name_mod = Option.value_exn ~message:err_msg_input
+             (get_z3_name init_var_map_mod reg Var.name) in
+         (* the pointer register must be above RSP
```
codyroux, line 141: As a side note, if you can get your hands on a module of type `CPU`, you can use this function: https://binaryanalysisplatform.github.io/bap/api/odoc/bap/Bap/Std/module-type-CPU/index.html#val-sp to get your hands on the actual variable that contains the top of the stack.

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -111,3 +111,59 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
+          *  NOTE: we are assuming stack grows down *)
+         let uge_1 = Z3.BitVector.mk_uge ctx reg_name_orig rsp_orig in
+         let uge_2 = Z3.BitVector.mk_uge ctx reg_name_mod rsp_mod in
+         (* the pointer must be below the end of the stack *)
+         let ule_1 =  Z3.BitVector.mk_ule ctx reg_name_orig sb_bv in
+         let ule_2 =  Z3.BitVector.mk_ule ctx reg_name_mod sb_bv in
+         (* encode that the pointer is either above RSP or below the stack
```
codyroux, line 148: Great! You might want to write this down as a formula, like `RSP <= R \/ R >= stack_end` or something. I find formulae to be slightly more readable in this context.

No big deal though.

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -111,3 +111,59 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
+          *  end and thereby outside the uninitialized stack region *)
+         let or_c_1 = Z3.Boolean.mk_or ctx [uge_1; ule_1] in
+         let or_c_2 = Z3.Boolean.mk_or ctx [uge_2; ule_2] in
+         let and_c = Z3.Boolean.mk_and ctx [or_c_1; or_c_2;] in
+         Z3.Boolean.mk_and ctx [and_c; acc]
+      )
+    | None ->
```
codyroux, line 155: Maybe worth commenting that this is the case where we're analyzing a single binary.

wp/plugin/lib/analysis.ml
```diff
@@ -138,6 +138,19 @@ let sp (arch : Arch.t) : (Comp.comparator * Comp.comparator) option =
            memory for the x86_64 architecture.\n%!";
     None
 
+
```
codyroux, line 141: Wait, why do we have a `lib` sub-folder in our `plugin` folder? Shouldn't this be in `wp/lib`?

wp/plugin/lib/analysis.ml
```diff
@@ -151,7 +164,8 @@ let comparators_of_flags
     post_reg_values p.compare_post_reg_values
       ~orig:(sub1, env1) ~modif:(sub2, env2);
     smtlib ~precond:p.precond ~postcond:p.postcond;
-    sp arch
+    sp arch;
```
codyroux, line 167: Hey there's the stack pointer!

wp/plugin/lib/analysis.ml
codyroux: Nevermind.

wp/resources/sample_binaries/pointer_flag/main_1.c
```diff
@@ -0,0 +1,15 @@
+int f(){
+    return 1;
+}
+// Type your code here, or load an example.
```
codyroux, line 4: What does this comment mean?

wp/resources/sample_binaries/pointer_flag/run_wp_sat.sh
```diff
@@ -0,0 +1,23 @@
+# This is a test that tests that the wp-pointer-reg-list flag works as expected.
+# Specifically, the function foo tries to use differing
```
codyroux, line 2: I see no function called `foo`...

wp/resources/sample_binaries/pointer_flag/run_wp_sat.sh
```diff
@@ -0,0 +1,23 @@
...
+# return addresses that are pushed onto the stack to cause a different return
+# value between functions. The abscence of the flag should allow the pointer
```
codyroux, line 4: It's spelled "absence". Not to worry, my spelling is atrocious, but you might want to run some quick spell checks, or enable the squigglies in your IDE.

wp/resources/sample_binaries/pointer_flag/run_wp_unsat.sh
```diff
@@ -0,0 +1,24 @@
+# This is a test that tests that the wp-pointer-reg-list flag works as expected.
```
codyroux, line 1: A test that tests the test.

wp/resources/sample_binaries/pointer_flag/run_wp_sat.sh
```diff
@@ -0,0 +1,23 @@
...
+# argument in RDI to be used to point to this return address, as it lives
+# on the uninitialized portion of the stack at the start of the function.
+
+# Should return SAT
```
codyroux, line 8: I actually don't get how this example is supposed to work. What's happening with `num1`? It might help to actually refer to the C names of variables and functions.

wp/resources/sample_binaries/pointer_flag/main_1.c
```diff
@@ -0,0 +1,15 @@
...
+int deref(int *num1) {
+    int num2 = f();
```
codyroux, line 6: you could have `int num2 = 12;` and then add a `int local = 42;` right after, that way you need to prove that `*num1` doesn't point to the local stack.

This is the main problem we're having anyway.

wp/lib/bap_wp/src/z3_utils.ml
DieracDelta: Yes! I added this into `environment.ml` as `get_rsp_name`. I realize the name is x86 specific, so I'm going to change the name.

DieracDelta:
I've made changes to incorporate in @codyroux 's suggestions. Specifically:
- Simplified the example in `pointer_flag`
- Various comment corrections
- Better error messages
- Fix a slight typo in the makefile

codyroux:
All clear from me!

PR #233, opened 2020-08-26
Fortunac/docs

fortunac:
Updated the site with installation instructions and the developer API, updated the tutorial and exercises with the new interface, and consolidated the flag documentation to the website,


wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
+            to check a single binary for certain behaviors.|}
 
 (* Mandatory arguments. *)
 
 let files = Cmd.arguments Typ.file
-    ~doc:"Path(s) to one or two binaries to analyze. If two binaries are \
-          specified, WP will run a comparative analysis."
+    ~doc:{|Path(s) to one or two binaries to analyze. If two binaries are
+           specified, WP will run a comparative analysis. If 0 or more than 2
+           files are give, WP will fail.|}
```
jtpaasch, line 30: This might be my personal preference, but the phrase "WP will fail" is ambiguous, and as a user, I think, "what does that mean? Will it fail silently? Will it throw an error? Will it give me false positives?" 

If it throws an error, I think it would be better to say "WP will throw an error".

wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
 
 let func = Cmd.parameter Typ.string ~aliases:["function"] "func"
-    ~doc:"Function to run the WP analysis on. If no function is specified or \
-          the function cannot be found in the binaries, the analysis will \
-          fail."
+    ~doc:{|Determines which function to verify. WP verifies a single function,
+           though calling it on the `main' function along with the `inline'
+           option will analyze the whole program. If no function is specified
+           or the function cannot be found in the binary/binaries, the analysis
+           will fail.|}
```
jtpaasch, line 37: Ditto about fail vs throw an error.

wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
 
 (* Arguments that determine which properties CBAT should analyze. *)
 
 let precond = Cmd.parameter Typ.string "precond"
-    ~doc:"A custom precondition in SMT-LIB format to be used during the \
-          analysis. If no precondition is specified, a trivial precondition \
-          of `true' will be used."
+    ~doc:{|Allows the introduction of assertions to the beginning of a query.
```
jtpaasch, line 42: I'm not sure a user knows what a "query" is (I don't know what that is, for example). I'm also not sure the user will understand what it means to prune possible models (I think I do know what that one means though). 

I wonder if it would be simpler to say something like "Allows you to specify an assertion that WP will assume is true at the beginning of the function it is analyzing. Assertions are specified in the smt-lib2 format..."

wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
+           expression. For example, `--precond="(assert (= RDI_mod
+           #x0000000000000003)) (assert (= RDI_orig #x0000000000000003))".' If
+           no precondition is specified, a trivial precondition of `true' will
+           be used.|}
 
 let postcond = Cmd.parameter Typ.string "postcond"
-    ~doc:"A custom postcondition in SMT-LIB format to be used during the \
-          analysis. If no postcondition is specified, a trivial \
-          postcondition of `true' will be used."
+    ~doc:{|Replaces the default postcondition with the user-specified one,
```
jtpaasch, line 53: I'm not sure the user knows there is a default postcondition that they need to replace. Similar to "precond," I wonder if it would be simpler to say something like "Allows you to specify an assertion that WP will assume is true at the end of the function it is analyzing, using the smt-lib2 format. Similar to ..."

wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
 
 let check_null_derefs = Cmd.flag "check-null-derefs"
-    ~doc:"If set, the WP analysis will check for inputs that would result in \
-          dereferencing a NULL value. In the case of a comparative analysis, \
-          WP will check that the modified binary has no additional null \
-          dereferences in comparison with the original binary."
+    ~doc:{|Checks for inputs that would result in dereferencing a null value
+           during a memory read or write. In the case of a comparative
+           analysis, checks that the modified binary has no additional null
+           dereferences in comparison with the original binary.|}
```
jtpaasch, line 67: I would say "In the case of comparative analysis, checks that the modified binary has no additional paths through the function that contain null dereferences in comparison with the original binary." It's the extra *paths* with null dereferences that WP finds, not extra null dereferences.

wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
-          compare the values stored in the specified registers at the end of \
-          the analyzed function's execution. For example, `RAX,RDI' compares \
-          the values of RAX and RDI at the end of execution. If unsure about \
-          which registers to compare, x86_64 architectures place their \
-          output in RAX, and ARM architectures place their output in R0."
+    ~doc:{|This flag is only used in a comparatve analysis. Compares the values
+           stored in the specified registers at the end of the function's
+           execution. For example, `RAX,RDI' compares the values of RAX and RDI
+           at the end of execution. If unsure about which registers to compare,
+           x86_64 architectures place their output in RAX, and ARM
```
jtpaasch, line 79: Maybe say "If unsure about which registers to compare, check the architecture's ABI. x86_64 architectures place their output in RAX, ..."

wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
+           inlined. When functions are not inlined, heuristic function
+           summaries are used at function call sites. For example, if you want
+           to inline the functions `foo' and `bar', you can write
+           `--inline=foo|bar'. To inline everything, use `--inline=.*' (not
+           generally recommended).|}
 
 let num_unroll = Cmd.parameter Typ.(some int) "num-unroll"
-    ~doc:"Amount of times to unroll each loop. By default, WP will unroll \
-          each loop 5 times."
+    ~doc:{|Replaces the default number of times to unroll each loop. WP will
```
jtpaasch, line 93: Instead of "Replaces the default number of times" maybe say "Specifies the number of times"

wp/plugin/wp.ml
```diff
@@ -18,149 +18,158 @@ module Typ = Extension.Type
...
 
 let mem_offset = Cmd.flag "mem-offset"
-    ~doc:"If set, maps the symbols in the data and bss sections from their \
-          addresses in the original binary to their addresses in the \
-          modified binary."
+    ~doc:{|This flag is only used in a comparative analysis. Maps the symbols
+           in the data and bss sections from their addresses in the original
+           binary to their addresses in the modified binary. If this flag is
+           not present, WP assumes that memory between both binaries are
+           equal.|}
```
jtpaasch, line 120: Instead of saying "assumes that memory between both binaries are equal", maybe say "assumes that memory between both binaries start at the same offsets"?

wp/plugin/wp.ml
fortunac: WP will print out an error message and then exit. I can replace this with "If 0 or more than 2 files are given, WP will not run an analysis and will exit."

wp/plugin/wp.ml
fortunac: Same as above. An error message will print and we exit. I can change this comment similarly.

wp/plugin/wp.ml
jtpaasch: I think "WP will exit with an error message" is just fine.

fortunac: Always coming up with the nice and concise documentation. :)

wp/plugin/wp.ml
fortunac: You're right. The default postcondition was changed to `true`, so there really is nothing to replace. This flag will just add postconditions on top of the ones the user specifies with flags.

fortunac:
> On line 238 in wp.ml, we should have some error handling instead of always returning Ok (). If the Analysis returns an error, then the callback should return an Error variant, otherwise Ok ().

I'm going to create a ticket for this change since it is a bit out of scope for this PR. Otherwise, all the requested changes, have been pushed.

docs/installation.html
```diff
@@ -0,0 +1,82 @@
...
+
+<a name="manual"></a>
+<h2>Manual Installation</h2>
+
+<h3>Prerequisites</h3>
+
+<p>Before installing, make sure you have the appropriate dependencies installed. First, there is OCaml:
+
+<ul>
+    <li>ocaml 4.07.0</li>
```
ccasin, line 40: We should specify here whether this specific version of ocaml is needed (and if not, which versions we support).  And also instructions for installing it, or at least a pointer to the website and a generic statement like "you can install ocaml through your system's package manager and install the correct switch with this command."

docs/installation.html
```diff
@@ -0,0 +1,82 @@
...
+    <li>bap 2.1.0</li>
+    <li>ounit 2.0.8</li>
+    <li>z3 4.8.8</li>
+    <li>re 1.9.0</li>
+</ul>
+
+<p>All of these can be installed with:
+
+<pre>
+  opam install package_name
```
ccasin, line 56: We should really have a command or sequence of commands that you can just copy/paste to install all the prereqs.

ccasin (approved):
This is great!  A couple minor suggestions, but you can make them a separate ticket and merge this now if you want.

fortunac:
> 
> 
> This is great! A couple minor suggestions, but you can make them a separate ticket and merge this now if you want.

Sounds good! These are small fixes, so I can add them to this PR.

PR #242, opened 2020-09-17
--pointer-reg-list: no longer error on registers that do not exist

DieracDelta:
#230 introduced a bug where including a register not in the analyzed sub would result in an error. For example, when invoking cbat with `--pointer-reg-list=ASDF` on any binary,  we would like to continue with analysis and ignore these extraneous registers that are not used. Additionally, the error was a string error, which should occur at the start/frontend instead of in the middle of analysis. (note: I'm mostly paraphrasing things @codyroux mentioned in slack) 

`construct_pointer_constraint` now takes in lists of `Constr.z3_expr`s instead of strings. The `string -> z3_expr` occurs much earlier on.

To continue with analysis and not throw an error, we do two things:

- Create the remaining registers in `--pointer-reg-list` right after the registers for the sub are initialized
- If the variable is somehow still not found, then issue a warning and continue with execution (this should never be triggered and exists more as a sanity check).

This should allow us to add in constraints for all registers provided (even if they do not exist or are unused in the sub). 

The tests *seem* to pass. I'm not 100% sure that I'm adding in the registers correctly. On the other hand, perhaps I'm being too careful. When adding in registers I:

1) Add all registers seen in the sub.
2) Call `init_vars` on those registers to update their init maps
3) Add in all registers listed in `pointer-reg-list` that don't already exist in the corresponding init map.
4) Call `init_vars` on the registers introduced in (3).

I'm not sure if this is overcomplicating things. Perhaps I should just `Var.create` on all the strings provided by `pointer-reg-list` and then union them with the set of variables found by `Pre.get_vars`?

### DIFF
--- a/wp/lib/bap_wp/src/z3_utils.ml
+++ b/wp/lib/bap_wp/src/z3_utils.ml
@@ -118,7 +118,7 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
   let arch = match Env.get_arch env1 |> Bap.Std.Arch.addr_size with
     | `r32 -> 32
     | `r64 -> 64 in
-  let rsp = Env.get_sp_name env1 in
+  let rsp = Env.get_sp env1 |> Var.name in
   let function_name = "construct_pointer_constraint: " in
   let err_msg_rsp = "stack pointer not found" in
   let sb_bv = Z3.BitVector.mk_numeral ctx (stack_end |> Int.to_string) arch in
--- a/wp/lib/bap_wp/src/z3_utils.ml
+++ b/wp/lib/bap_wp/src/z3_utils.ml
@@ -111,44 +111,40 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
   let ctx = Env.get_context env in
   mk_smtlib2 ctx smt_post decl_syms
 
-let construct_pointer_constraint (l: string list) (env1 : Env.t)
-    (env2: Env.t option) : Constr.t =
+let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
+    (l_mod : (Constr.z3_expr list) option) (env2: Env.t option) : Constr.t =
   let stack_end = Env.get_stack_end env1 in
   let ctx = Env.get_context env1 in
   let arch = match Env.get_arch env1 |> Bap.Std.Arch.addr_size with
     | `r32 -> 32
     | `r64 -> 64 in
   let rsp = Env.get_sp_name env1 in
-  let function_name = "construct_pointer_constraint" in
-  let err_msg_input = "invalid register name in pointer register list" in
+  let function_name = "construct_pointer_constraint: " in
   let err_msg_rsp = "stack pointer not found" in
   let sb_bv = Z3.BitVector.mk_numeral ctx (stack_end |> Int.to_string) arch in
-  let gen_constr = match env2 with
+  let gen_constr, l = match l_mod, env2 with
     (* comparative case *)
-    | Some env2 ->
+    | Some l_mod, Some env2 ->
       let init_var_map_orig = Env.get_init_var_map env1 in
       let init_var_map_mod = Env.get_init_var_map env2 in
-      (* we do want exceptions here if the register names are invalid
-       *  or RSP doesn't exist *)
-      let rsp_orig = Option.value_exn ~message:(function_name ^ ": original " ^ err_msg_rsp)
+      (* we do want exceptions here if RSP doesn't exist *)
+      let rsp_orig = Option.value_exn
+          ~message:(function_name ^ "original " ^ err_msg_rsp)
           (get_z3_name init_var_map_orig rsp Var.name) in
-      let rsp_mod = Option.value_exn ~message:(function_name ^ ": modified " ^ err_msg_rsp)
+      let rsp_mod = Option.value_exn
+          ~message:(function_name ^ "modified " ^ err_msg_rsp)
           (get_z3_name init_var_map_mod rsp Var.name) in
       (* Encode constraint that each register is not within stack*)
-      (fun acc reg ->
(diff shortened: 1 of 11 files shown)


fortunac:
>  When adding in registers I:
>     1. Add all registers seen in the sub.
>     2. Call `init_vars` on those registers to update their init maps
>     3. Add in all registers listed in `pointer-reg-list` that don't already exist in the corresponding init map.
>     4. Call `init_vars` on the registers introduced in (3).

What was your reasoning for calling `init_vars` twice? Does creating a set that contains all the registers found in the sub and all the registers listed in `pointer-reg-list`, and then passing that to `init_vars` not work?

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -111,44 +111,40 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
     (* comparative case *)
-    | Some env2 ->
+    | Some l_mod, Some env2 ->
       let init_var_map_orig = Env.get_init_var_map env1 in
       let init_var_map_mod = Env.get_init_var_map env2 in
-      (* we do want exceptions here if the register names are invalid
-       *  or RSP doesn't exist *)
-      let rsp_orig = Option.value_exn ~message:(function_name ^ ": original " ^ err_msg_rsp)
+      (* we do want exceptions here if RSP doesn't exist *)
+      let rsp_orig = Option.value_exn
```
fortunac, line 131: Does this error here if I have a function that doesn't use RSP? For example, something like

```
sub():
RDI := 0xDEADBEEF
call RDI with noreturn
```

DieracDelta: Yes, I just tried and this does error. Perhaps the right thing to do here is to not error, generate an `true` constraint, and print a warning to the log.

fortunac: What do you do for other registers that can't be found in the subroutine?

DieracDelta: For the other registers, if they do not exist, I create them, create their respective constraint (which ends up being trivial since they are unused in the sub), and print a warning to the log. To be consistent, I can also do this with the stack pointer.

fortunac: Since RSP is used in the constraint generated, if we do create this trivial constraint, is it equivalent to true? I'm actually having a hard time trying to figure out which is the final constraint generated from the comments. Can you enlighten me haha?

wp/plugin/lib/analysis.ml
```diff
@@ -138,19 +138,55 @@ let sp (arch : Arch.t) : (Comp.comparator * Comp.comparator) option =
            memory for the x86_64 architecture.\n%!";
     None
 
+(* Helper to look up the variable name in the initial variable map. *)
+let get_init_z3_expr (reg : string) (env: Env.t) : Constr.z3_expr option =
```
fortunac, line 142: Is there a reason why we don't use `Env.get_init_var`? https://github.com/draperlaboratory/cbat_tools/blob/6005858ffe338d00348e1eacc316c7a25852bd2b/wp/lib/bap_wp/src/environment.ml#L481

wp/lib/bap_wp/src/z3_utils.ml
DieracDelta: Suppose, for some register `R`, we generate this constraint on a single binary (the comparative case is analogous). Then we would have the constraint be:

```
R >= RSP \/ R <= stack_bottom
```
Since `RSP` can be chosen to be anything (because it's not used in the sub), `CBAT` can choose it to be `stack_bottom`. I think this evaluates to `true` in that case. The final constraint is a bunch of these constraints (`R` is the only thing that changes) anded together, which should remain as true.

fortunac: Okay! And `RSP <= RSP` is true, so I'm convinced this is true. Go ahead and place it with the other variables to be more consistent and add a big warning log for this.

wp/plugin/lib/analysis.ml
DieracDelta: The issue I ran into was that I couldn't find a simple way to get a variable from a string without creating a new set of variables. I've  refactored; now I just pass in the created variables from the single/comparative analysis and use this function. This feels much cleaner!

DieracDelta:
You're right--I don't think there is a good reason to have two sequential invocations to `init_vars`. I was mostly using this as a convenience to generate errors. I've refactored to only have one invocation over a union of the two sets and this seems to work well!

wp/plugin/lib/analysis.ml
```diff
@@ -138,25 +138,48 @@ let sp (arch : Arch.t) : (Comp.comparator * Comp.comparator) option =
            memory for the x86_64 architecture.\n%!";
     None
 
+(* Create list of variables from string names. *)
+let create_vars (l : string list) : Bap.Std.Var.Set.t =
+  List.map l ~f:(fun var -> Var.create var reg64_t) |> Bap.Std.Var.Set.of_list
```
fortunac, line 143: Hardcoding the variables as `reg64_t` kinda worries me, especially in the case where our input registers happen to be 32-bit. Is there a way to determine the width of the register? Perhaps the Arch or Target modules have something? I'm not really sure.

DieracDelta: There's nothing as far as I can see in `Arch` or `Target` that fetches the register size. The two solutions I see are:
- Use the width of the sp (which I can snag off of `Target`) as the width
- Use the word size (which I can grab from `Arch`) as the width.

I'm not sure which would be preferable here.

DieracDelta: Following @fortunac 's [observation](http://binaryanalysisplatform.github.io/bap/api/master/bap-x86-cpu/X86_cpu/AMD64/index.html) of `Target.CPU.gpr`, I've updated `create_vars` to either return a general purpose register if its name matches the input, or fail if the name is not in the set of general purpose registers for the program's architecture.

PR #243, opened 2020-09-18
Update position independent run script

fortunac:


philzook58:
Seems simple enough. Tests run and pass on my system except for "Goto String" and "Hash Function", which we're aware of

Issue #227, opened 2020-08-09
Try other SMT solvers


philzook58:
Since we have added the capability to print the smtlib2 query, it is not so hard to evaluate other SMT solvers (in particular CVC4 and Boolector). If there is significant variation among solvers, it may be worth running them in a parallel portfolio in separate processes. We don't necessarily even have to replace our usage of Z3 for constraint construction to achieve this.

philzook58:
Yices 2 is also promising seeming 
https://smt-comp.github.io/2020/results/qf-aufbv-single-query

philzook58:
On a very preliminary experiment for a particular query (addname in grammatech)
- z3 3m20s
- yices 54s
- cvc4 timeout
- boolector 15s

philzook58:
Here is a rough outline of a relatively painless way to integrate other solvers.
Add a flag to point to a smtlib compatible binary `--wp-solver=~/Downloads/boolector/bin/boolector`

Fork a process calling this binary name in `Precondition.check`. I'm not sure exactly how to do this. core probably has this functionality. Maybe this https://ocaml.janestreet.com/ocaml-core/latest/doc/core/Core__/Core_command/index.html .   Roughly you need to pipe in the following string to the process:
````
Printf.printf "(set-logic QF_AUFBV) \n %s \n (check-sat) \n (get-model) %!" (Z3.Solver.to_string solver) );
```
You may want to check that it returns "sat" before asking for the model. I think solvers can often be run interactively via pipe.

The model should look something like this http://smtlib.cs.uiowa.edu/examples.shtml
```
(get-model)
; ((define-fun x () Int 8)
;  (define-fun y () Int 6)
; )
```
Take this, parse as an SExpr and then assert these values to the z3 solver and call Z3.Solver.check if sat.

codyroux:
@philzook58 can we close this?

philzook58:
I suppose we can. This work has been integrated into master

PR #244, opened 2020-09-18
Phil/ext smt

philzook58:
Adds the ability to call an external smt solver with new option `--ext-solver-path`. Only tested with Boolector. https://boolector.github.io/
Includes some changes to creation of z3 query for smtlib compatibility, specifically removal of single and no argument conjunction, and replacing bv-red-or with ite equivalent.

`Z3_utils.check_external` invokes an external smt solver as a process. It communicates to the 
    process via a fairly rigid interpretation of the smtlib2 protocol. It extracts the 
    smtlib2 query string from the given z3_solver, queries the external solver, receives 
    a counter model if SAT. It then asserts this model back to the z3_solver, which should
    be able to check quickly. This solver can then be used with the regular cbat z3 model 
    interpretation machinery. Tested with Boolector, results with other solvers may vary.*


wp/lib/bap_wp/src/constraint.ml
```diff
@@ -246,12 +246,12 @@ let rec eval_aux ?stats:(stats = init_stats) (constr : t) (olds : z3_expr list)
       (eval_aux ~stats:stats c2 olds news ctx)
   | Clause (hyps, concs) ->
     let eval_conjunction conj =
-      if List.length conj = 1 then 
-        (* This is tail recursive. Avoids And node. *)
-        eval_aux ~stats:stats (List.hd_exn conj) olds news ctx 
-      else
+      (* This adds possible tail recursiion, avoids And node. *)
```
codyroux, line 249: "adds possible"? Is it tail recursive or not?

Also: `recursiion`.

philzook58: Some branches are tail recursive, some aren't. Depends on the input

philzook58:
This includes the changes from #229

philzook58:
A note:
Currently there are some variables that are not transferred from the Boolector model over to Z3. In particular I've noticed that the chaos return values of functions are not included. When you import smtlib into z3, you need to explicitly list the already defined variables and there is not a way to enumerate them from a Z3.solver. I don't know if we currently store these in the env is the only obstruction from including them. Z3 has the ability to fill in the gaps. The more gaps z3 has to fill, the slower it might be. 

Other todos:
- run and make sure the boolector and z3 version agree on sat VS unsat for all queries

philzook58:
Related to the discoveries documented in issue #227

philzook58:
Preliminary results:
image (6)
A graph of relative time between z3 and boolector for the first 130 functions in G.

Functions that performed >20s worse with boolector. Solve time difference in seconds. 
```
name
_flush_write            -514.210
_wsplt_nomem            -136.997
add_literal_segment      -94.415
areadlinkat_with_size   -480.175
Name: JobRuntime, dtype: float64
```

Functions where boolector did better by >20s.
```
_argp_fmtstream_ensure.part.0    229.520
_obstack_newchunk                795.298
_wsplt_error                     913.792
acls_one_line.constprop.4        405.478
add_chunk_header.isra.3          934.622
add_exclude_fp                   636.647
add_exclusion_tag                461.994
add_hierarchy_to_namelist        169.848
add_starting_file                754.914
add_volume_label                  61.024
addname                          391.766
all_names_found                   76.046
alloc_action                     148.800
alloc_space                       95.232
argmatch_valid                   484.251
argp_default_parser              113.549
argp_doc                         553.556
argp_fmtstream_printf            815.915
assign_time_option                97.389
atime_decoder                     21.338
bufmap_free                      202.548
```

For the functions that timeout at 1000s for z3, this is the relative performance for boolector. (large positive numbers = no longer timing out)
```
_argp_fmtstream_update                 -0.234
_getopt_internal_r                     -0.084
_help                                  -0.212
_obstack_newchunk                     795.298
_wsplt_error                          913.792
_wsplt_subsplit                        -1.021
acls_one_line.constprop.4             405.478
add_chunk_header.isra.3               934.622
add_exclude                             0.832
add_hierarchy_to_namelist             169.848
add_starting_file                     754.914
addname                               391.766
append_incremental_renames              0.223
apply_delayed_links                    -0.004
apply_nonancestor_delayed_set_stat     -1.219
argp_args_usage                        -0.018
argp_doc                              553.556
argp_error                              0.408
argp_failure                            0.550
argp_fmtstream_printf                 815.915
```

Conclusion: For most functions about the same, but for a large fraction Boolector is significantly better.

philzook58:
On a complete test of G*********, 
```
z3 sat: 582 unsat: 388 unknown: 202
boolector sat: 651 unsat: 393 unknown: 128
```
There were no inconsistent values found (sat returning unsat or unsat returning sat). It was not a clean sweep. Boolector returned unknown for some problems Z3 returned on, although few.

```
z3 sat, boolector unknown : {'listxattrat', 'argp_make_fmtstream', 'human_options', 'hash_initialize', 'xheader_set_option', z3 sat, 'dumpdir_create0', 'llistxattrat'}

z3 unsat, boolector unknown :  {'quotearg_alloc_mem'}

z3 unknown, boolector sat : {'xheader_decode', 'extract_fifo', 're_node_set_insert', 'extract_node', 'hash_find_entry', 'xheader_finish', 'peek_token', 'short_read', 'name_gather', 'sparse_add_map', 'duplicate_node_closure', 'argmatch_valid', 'decode_timespec', 'create_archive', 'mode_adjust', 'wordsplit_add_segm', 'read_obstack', 'lower_subexps', 'wsnode_nullelim', 'owner_map_translate', 'read_num', 'check_data_region', 'mktime_z', 'duplicate_tree', 'dbg_printf', 'xheader_string_end', 'write_directory_file_entry', 'argp_fmtstream_printf', 'sparse_scan_file_raw', 'add_starting_file', 'hash_delete', 'build_upper_buffer', 're_node_set_init_union', 'write_gnu_long_link', 'group_map_translate', '_obstack_newchunk', 'merge_sort_sll', '_wsplt_error', 'sparse_extract_file', 'match_ctx_clean', 'sparse_dump_region', 'hash_clear', 'exchange', 'wordsplit_tildexpand', 'acls_one_line.constprop.4', 'add_exclude_fp', 'register_state', 'file_count_links.part.4', 'sys_exec_info_script', 'remove_delayed_set_stat', 'transform_stat_info', 'extend_buffers', 'xasprintf', 'write_directory_file', 'pop_fail_stack.isra.25', 'check_arrival_expand_ecl', 'chdir_long', 'check_dst_limits_calc_pos_1', 'create_token_tree.isra.16', 'chdir_do', 'check_time', 'hash_rehash', '__getdelim', 'to_chars', 'check_subexp_matching_top.isra.18', 'xheader_print_n', 'name_list_advance', 'extract_link', 'optloc_save', 'sparse_extract_region', 'expcmd', 're_acquire_state', 'sub_epsilon_src_nodes', 'get_subexp_sub.isra.32', 'report_difference', 'info_attach_exclist'}

z3 unknown, boolector unsat {'opendir_safer', 'revert_tz.part.1', 'add_chunk_header.isra.3', 'addname', 'finish_header', 'rpl_fcntl'}

```

Runtime with 1000s timeout and -j10 parallelism
z3 : 7.7 hours
boolector: 5.1 hours

wp/lib/bap_wp/src/environment.ml
```diff
@@ -495,7 +502,7 @@ let mk_init_var (env : t) (var : Var.t) : Constr.z3_expr * t =
   let ctx = get_context env in
   let z3_var, _ = get_var env var in
   let sort = Expr.get_sort z3_var in
-  let name = Format.sprintf "init_%s" (Expr.to_string z3_var) in
+  let name = Format.sprintf "init_%s" (String.strip ~drop:(fun c -> Char.(c = '|')) (Expr.to_string z3_var)) in
```
fortunac, line 505: Can you explain what this line is for?

philzook58: Z3 (and smtlib) has a quoting mechanism where if a name contains certain characters it will wrap the name in "|" for example init_#248 will be printed as |init_#248| because # is a special character. It does this when you print Expr, but not when you print the Symbols from Z3 or make them. Since I'm taking the names from Expr to build the symbols, I'm stripping off this "|" character. This does deserve a comment, so I was also confused what this was doing exactly when i came back to it.

wp/lib/bap_wp/src/output.ml
```diff
@@ -37,31 +37,42 @@ let format_mem_model (fmt : Format.formatter) (mem_model : mem_model) : unit =
...
-    end
-    else if Int.(numargs = 1) && String.(f_name = "const") then begin
-      let key = List.nth_exn args 0 in
-      { default = key; model = List.rev partial_map }
-    end
-    else begin
-      warning "Unexpected case destructing Z3 array: %s" (Z3.Expr.to_string e);
-      { default = e ; model = partial_map }
+    (* FIXME: Better condition for detecting lambda term. *)
+    if numargs <= 3 then begin
```
fortunac, line 51: I haven't tested it, but does `Z3Array.is_array` not work here?

wp/lib/bap_wp/src/environment.ml
fortunac: So what's happening is that the `|`'s cause things to break while using a different SMT solver when trying to create symbols based off of the expr strings?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1302,9 +1309,9 @@ let mem_read_offsets (env2 : Env.t) (offset : Constr.z3_expr -> Constr.z3_expr)
   if List.is_empty conds then
     None
   else
-    Some (Assume (AfterExec (Constr.mk_goal name (Bool.mk_and ctx conds))))
+    Some (Assume (AfterExec (Constr.mk_goal name (Z3_utils.mk_and ctx conds))))
 
-let check ?refute:(refute = true) ?(print_constr = []) ?(debug = false)
+let check ?refute:(refute = true) ?(print_constr = []) ?(debug = false) ?ext_solver
```
fortunac, line 1314: Can we add the default value for `ext_solver` to be more consistent with the other arguments in this function?

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -51,7 +51,7 @@ val z3_expr_zero : Z3.context -> int -> Constr.z3_expr
 val z3_expr_one : Z3.context -> int -> Constr.z3_expr
 
 (** Translate a BIR binary operator to a Z3 one. *)
-val binop : Z3.context -> Bap.Std.binop -> Constr.z3_expr -> Constr.z3_expr -> Constr.z3_expr
+val binop : Env.t -> Bap.Std.binop -> Constr.z3_expr -> Constr.z3_expr -> Constr.z3_expr
 
 (** Translate a BIR unary operator to a Z3 one. *)
 val unop : Z3.context -> Bap.Std.unop -> Constr.z3_expr -> Constr.z3_expr
```
fortunac, line 57: Not super important, but what are your thoughts on changing the argument from `context` to `env` in `unop` so that it continues to mirror the api for `binop`?

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -110,3 +110,160 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
+            model_val
+          | Sexp.List _  ->  (* Function model *)
+            Sexp.List [Sexp.Atom "lambda" ; args; model_val] (* Sexp.Atom varname *)
+          | Sexp.Atom _ -> failwith "Unexpected atom"
+        in
+        Sexp.List [Sexp.Atom "assert" ;
+                   Sexp.List [Sexp.Atom "=" ; Sexp.Atom varname ; model_val]]
+      else
+        begin
+          Printf.printf "Warning: %s not instantiated in Z3 query\n" varname;
```
fortunac, line 175: Would it be better to make this a `warning` log?

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -110,3 +110,160 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
+    [and] node if the number of operands is less than 2. This may improve sharing,
+    but also improves compatibility of smtlib2 expressions with other solvers  *)
+let mk_and ( ctx : Z3.context ) (xs : Constr.z3_expr list) : Constr.z3_expr =
+  match xs with
+  | []  -> Bool.mk_true ctx
+  | [x] -> x
+  | _   -> Bool.mk_and ctx xs
+
+
+(** [asserts_of_model] takes a string of an smtlib2 model returned by an external solver and
```
fortunac, line 125: This is a really nice explanation!

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -110,3 +110,160 @@ let mk_smtlib2_single (env : Env.t) (smt_post : string) : Constr.t =
...
+  (* Send query to solver *)
+  Out_channel.output_string solver_stdin smt_preamble;
+  Out_channel.output_string solver_stdin smt_string;
+  Out_channel.output_string solver_stdin smt_postamble;
+  Out_channel.flush solver_stdin;
+  printf "Running external solver %s\n%!" solver_path;
+
+  (* SexpLib unfortunately uses # as an comment delimitter.
+     We replace it with a special token and revert this after parsing. *)
+  let pound_token = "MYSPECIALPOUND682" in
```
fortunac, line 213: Does this string have any special meaning? Or is it an arbitrary name only used for parsing?

wp/lib/bap_wp/src/output.ml
philzook58: That is a very good idea. It might not, but I should try. I couldn't find an is_lambda function

wp/lib/bap_wp/src/precondition.ml
philzook58: We can, but the default value is None. Is that what you mean?

fortunac:
This is very impressive work! Would it be possible to add to the documentation the following information?

- information about the new plugin flag
- possibly a list of smt solvers we support and a link to them

Would it also be possible to add a few tests that use Boolector and set up ounit to skip them if Boolector isn't installed?

wp/lib/bap_wp/src/z3_utils.ml
philzook58: No it has no special meaning. This extremely embarrassing section is a hack to try and work around the incompatibility of ocaml Sexp and smtlib

wp/lib/bap_wp/src/environment.ml
philzook58: It's really that internally to Z3 the reader for Symbol and the printer for Expr are not compatible in the presence of quotation I believe. That sounds crazy. Maybe I should double check that.

wp/lib/bap_wp/src/precondition.ml
fortunac: Yup! The optional arguments are all written as `?(arg = default)`, so I think changing to `?(refute = true)` and `?(ext_solver = None)` would be more consistent.

wp/lib/bap_wp/src/output.ml
philzook58: Yes this worked. Nice.

wp/lib/bap_wp/src/precondition.mli
philzook58: Hmm. I remember this being annoying and unsatisfying. I think maybe passing in context and an optional smt-compat bool flag might be better. Passing in the env is overkill

wp/lib/bap_wp/src/environment.ml
fortunac: I think that makes sense. It might be a good idea to ping @gltrost about this since it seems like he is doing some handling of the `|`s found in our queries.

wp/lib/bap_wp/src/precondition.mli
fortunac: Sounds good! We can leave this as is.

wp/lib/bap_wp/src/environment.ml
gltrost: Thanks for the ping. Yes, this is related to issue #300 I'm working on, and dealing with the `#`-vars as they cause escaped `\|`s will make printing more readable.

wp/lib/bap_wp/src/z3_utils.ml
fortunac: I can't think of a better way to handle this. I think the comment is sufficient enough for now.

wp/lib/bap_wp/src/precondition.ml
philzook58: Hmm. Actually I think I'm using a slightly different optional keyword argument mechanism here that for the other arguments. It doesn't accept an explicit (ext_solver = None).

wp/lib/bap_wp/src/output.ml
philzook58: Hmm. This goofs a little with the output when in Z3 mode
```
        mem_orig |-> [
                #x000105fc |-> 0x20 ;
                #x000105fd |-> 0x20 ;
                #x000105fe |-> 0x40 ;
                #x000105ff |-> 0x00 ;
                #x00402020 |-> 0x40 ;
                #x00402021 |-> 0x00 ;
                #x00402022 |-> 0x00 ;
                #x00402023 |-> 0x00 ;
                else |-> #x10]
```

fortunac: I'm a bit confused here. What changed?

philzook58: else |-> #x10] I think this used to be  `else |-> 0x10]` 
I suspect the constant case is not triggering is_array check so it's going through the other print statement

philzook58: We could use Expr.to_string on everything rather than Expr.expr_to_hex. Then I could remove the if statement also

fortunac: Okay, I see it now! Not sure what's best, so I'll leave it to your judgment here. I do have a slight preference for `0x10` over `#x10` though.

fortunac:
Integration tests are passing, and the updated documentation looks good!

philzook58:
I fixed the printing of the else case. AFAIK there are no other outstanding issues

fortunac:
> 
> 
> I fixed the printing of the else case. AFAIK there are no other outstanding issues

Would the changes here have any significant impacts on #300, @gltrost ?
https://github.com/draperlaboratory/cbat_tools/blob/f92c3cf9c65a50b15dfd6fd26f50702c84bb8acb/wp/lib/bap_wp/src/environment.ml#L505

gltrost:
> > I fixed the printing of the else case. AFAIK there are no other outstanding issues
> 
> Would the changes here have any significant impacts on #300, @gltrost ?
> https://github.com/draperlaboratory/cbat_tools/blob/f92c3cf9c65a50b15dfd6fd26f50702c84bb8acb/wp/lib/bap_wp/src/environment.ml#L505

I think this will minorly, but positively impact #300, since this chance addresses escaping `|`-symbols which was one of the issues in #300.

fortunac (approved):
Looks good to me! Good work!

Issue #245, opened 2020-09-23
Parallelize internal to WP


philzook58:
For whole programs, currently we analyze by running the wp on each function in the binary. This is embarrassingly parallel, and we use gnu parallel to speed it up. Unfortunately, the boot up overhead from loading the bap cache is significant when one is running 1000+ functions. On one example that starts with a G, every load took 50s, when the analysis often only took ~10s.

One way to fix this is to move some of the loop inside of the wp codebase, such that we only have to load the cache once. It is not obvious to me that we can parallelize entirely inside an ocaml program without using unix processes.

A suggested prcoessing flow is to
- Boot ~10 instances of wp (~one per core, or as many as your RAM can withstand)
- each instance of the wp reads from a named FIFO pipe or uses gnu parallel pipe https://www.gnu.org/software/parallel/parallel_tutorial.html#pipe as a queue of function names. `cat names.txt | parallel --pipe -j10 --roundrobin bap wp yada yada --stdin-loop``
- Each process runs it's analysis on the function name read from an In_channel in a loop. It uses a timeout for Z3 and a timeout wrapped around the entire analysis. https://ocsigen.org/lwt/3.2.1/api/Lwt_timeout

ivg:
So am I right that for each N function in a binary you load this binary N times? Why don't you just load it once, then fork M times after the KB is loaded and run in M jobs?

philzook58:
That also makes sense and might be a better road than what I was suggesting. I think we can't eagerly fork the entirety of the processes at once, so we'd need to throttle how many happen. It feels to me like we'd have to implement some percentage of a parallel queue system, unless there is a handy ocaml package that already does so (async-parallel? lwt?). This isn't impossible, but GNU parallel was already working pretty well and was very simple.

ivg:
Well yes, I understand that there is a big difference between using gnu parallel and implementing your own parallel execution framework :) And yes, there are a few libraries for parallel execution, e.g., parmap, forkwork, and even lwt-parallel by myself for that. But this will require significant rewriting of the wp plugin (or writing another plugin).

I am not sure what in particular you're doing in wp, but if your analysis is absolutely pure (i.e., you just take the sub and build wp and do not run any KB related stuff, in particular, you do not create any new terms) then you can just store each subroutine in a different file and run your analysis on it. 

To minimize the impact on wp code, which most likely requires `project` or `program` structures, you can create your own loader, which takes the empty project and puts a program with a single program into it.

PR #248, opened 2020-09-24
Save a pruned version of the kb to the cache

### LINKED ISSUE #247: architecture error
I have three binaries compiled for x8664. When I invoke them consecutively as below, the third invocation in the following error:

```
 bap wp --func=branching --trip-asserts --num-unroll=0 -- bin/main_6
 bap wp --func=branching --trip-asserts --num-unroll=0 -- bin/main_7
 bap wp --func=branching --trip-asserts --num-unroll=0 -- bin/main_8
```

```
Uncaught exception:

  "Option.value_exn None"

Raised at file "src/error.ml" (inlined), line 9, characters 14-30
Called from file "src/option.ml", line 69, characters 4-21
Called from file "lib/analysis.ml", line 183, characters 13-52
Called from file "lib/analysis.ml", line 298, characters 4-40
Called from file "wp.ml", line 249, characters 2-32
Called from file "cmdliner_term.ml", line 25, characters 19-24
Called from file "cmdliner_term.ml", line 23, characters 12-19
Called from file "cmdliner.ml", line 117, characters 32-39
Called from file "cmdliner.ml", line 147, characters 18-36
Called from file "cmdliner.ml", line 265, characters 22-48
Called from file "lib/bap_main/bap_main.ml", line 1054, characters 10-99
Called from file "lib/bap_main/bap_main.ml", line 1207, characters 15-140
Called from file "src/bap_frontend.ml", line 306, characters 8-127
```

This seems to be an error in determining the architecture (comes back as `None` instead of x8664). As suggested by Chloe, the quick way around this is `bap cache --clean`.

The binaries are attached below. Note that I had to rename them to `gif`s, though they are not actually gifs.
main_6
main_7
main_8

fortunac:
Based off of our discussion yesterday, this PR tests saving a pruned version of the kb to the cache. On some functions (tar_dirname), this changed runtime from 33s to 9s, but on others, (parse_branch), this changed runtime from 1m46s to 2m45s. I'm not really sure if this is the approach we want to go for.

This also might fix #247. I wasn't able to reproduce this bug unfortunately, but I had to update some parts of the architecture retrieval API. If you could test this for me, that would be great!

### DIFF
--- a/wp/plugin/lib/cache.ml
+++ b/wp/plugin/lib/cache.ml
@@ -16,6 +16,7 @@ open Bap_main
 open Bap_core_theory
 open Bap.Std
 open Regular.Std
+open KB.Let_syntax
 
 module Digests = struct
 
@@ -87,8 +88,6 @@ end
 
 module Program = struct
 
-  open KB.Let_syntax
-
   let package = "program"
 
   (* Gives a unique name to a KB object from the filename of the program. *)
@@ -150,6 +149,21 @@ module Arch = struct
       |> Option.bind ~f:Arch.of_string
     | Error c ->
       let msg = Format.asprintf "Unable to collect arch: %a%!"
+          KB.Conflict.pp c in
+      failwith msg
+
+  (* Adds the program's architecture to the KB. *)
+  let save file arch : unit =
+    let obj =
+      Theory.Unit.for_file file >>= fun label ->
+      let arch = Arch.to_string arch in
+      KB.provide Theory.Unit.Target.arch label (Some arch) >>| fun () ->
+      label in
+    let state = Toplevel.current () in
+    match KB.run Theory.Unit.cls obj state with
+    | Ok (_, state) -> Toplevel.set state
+    | Error c ->
+      let msg = Format.asprintf "Unable to provide architecture: %a%!"
           KB.Conflict.pp c in
       failwith msg
 
--- a/wp/plugin/lib/cache.ml
+++ b/wp/plugin/lib/cache.ml
@@ -23,7 +23,8 @@ module Digests = struct
   module Conf = Extension.Configuration
 
   (* Returns a function that makes digests. *)
-  let get_generator ctx ~filepath ~loader =
+  let get_generator (ctx : ctxt) ~(filepath : string) ~(loader : string)
+    : namespace:string -> digest =
     let inputs = [Conf.digest ctx; Caml.Digest.file filepath; loader] in
     let subject = String.concat inputs in
     fun ~namespace ->
@@ -31,30 +32,32 @@ module Digests = struct
       Data.Cache.Digest.add d "%s" subject
 
   (* Creates a digest for the knowledge cache. *)
-  let knowledge mk_digest = mk_digest ~namespace:"knowledge"
+  let knowledge (mk_digest : namespace:string -> digest) : digest =
+    mk_digest ~namespace:"knowledge"
 
   (* Creates a digest for the project state cache. *)
-  let project mk_digest = mk_digest ~namespace:"project"
+  let project (mk_digest : namespace:string -> digest) : digest =
+    mk_digest ~namespace:"project"
 
 end
 
 module Knowledge = struct
 
   (* Creates the knowledge cache. *)
-  let knowledge_cache () =
+  let knowledge_cache () : KB.state Data.Cache.t =
     let reader = Data.Read.create ~of_bigstring:KB.of_bigstring () in
     let writer = Data.Write.create ~to_bigstring:KB.to_bigstring () in
     Data.Cache.Service.request reader writer
 
   (* Loads knowledge (if any) from the knowledge cache. *)
-  let load digest : unit =
+  let load (digest : digest) : unit =
     let cache = knowledge_cache () in
     match Data.Cache.load cache digest with
     | None -> ()
     | Some state -> Toplevel.set state
 
   (* Saves knowledge in the knowledge cache. *)
-  let save digest : unit =
+  let save (digest : digest) : unit =
     let cache = knowledge_cache () in
     Toplevel.current ()
     |> Data.Cache.save cache digest
@@ -64,7 +67,7 @@ end
 module Project = struct
 
   (* Creates a project state cache. *)
-  let project_cache () =
+  let project_cache () : Project.state Data.Cache.t =
     let module State = struct
       type t = Project.state [@@deriving bin_io]
     end in
@@ -75,12 +78,12 @@ module Project = struct
     Data.Cache.Service.request reader writer
 
   (* Loads project state (if any) from the cache. *)
-  let load digest : Project.state option =
+  let load (digest : digest) : Project.state option =
     let cache = project_cache () in
     Data.Cache.load cache digest
 
   (* Saves project state in the cache. *)
-  let save digest state : unit =
+  let save (digest : digest) (state : Project.state) : unit =
     let cache = project_cache () in
     Data.Cache.save cache digest state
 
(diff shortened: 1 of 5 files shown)


wp/plugin/lib/cache.ml
```diff
@@ -153,5 +152,20 @@ module Arch = struct
           KB.Conflict.pp c in
       failwith msg
 
+  (* Adds the program's architecture to the KB. *)
+  let save file arch : unit =
```
philzook58, line 156: It would be nice to add types here for consistency and readability

fortunac: I can definitely add them! It looks like I failed to add the types for the functions in this entire file, so I can add those too to be consistent.

DieracDelta:
This looks good to me and fixes issue #247 for me!

DieracDelta (approved):

wp/plugin/lib/cache.ml
```diff
@@ -74,33 +78,32 @@ module Project = struct
...
 
 module Program = struct
 
-  open KB.Let_syntax
-
   let package = "program"
 
   (* Gives a unique name to a KB object from the filename of the program. *)
-  let for_program filename =
+  let for_program (filename : string) : Theory.Unit.cls KB.obj KB.t =
```
ivg, line 97: A package is a unit of administrative control, i.e., it is a namespace that is controlled by an organization or a sub-organization. You should take the package names that correspond to your group or your project, e.g., `draper`, `bap`, or `vibes`. When you need sub-organizations, sub-teams, or sub-projects, use sub-packages, e.g., `draper.vibes` or even `edu.cmu.ece.bap`, cf Java packages. CC @jtpaasch

fortunac: Thanks for letting us know! I'll update this right away.

ivg: On the other hand, we also use packages like `file` in the knowledge base, but for objects, so here the package `program` fits nicely in the concept, but using the same package for the slot is definitely a no-no. Not only it tautological, i.e., you have a property `program:program` of the class `core-theory:program` but it also hinders composability as you program will not work if someone else will decide to add the same named property to the class program.

wp/plugin/lib/cache.ml
```diff
@@ -153,5 +156,20 @@ module Arch = struct
           KB.Conflict.pp c in
       failwith msg
 
+  (* Adds the program's architecture to the KB. *)
+  let save (file : string) (arch : Arch.t) : unit =
```
ivg, line 160: This will break once #1225 is merged. Also, I am not sure I understand why is this needed, can you clarify?

wp/plugin/lib/cache.ml
fortunac: Would the package `draper.cbat.wp` work? With this, we can have the property, `draper.cbat.wp:program`.

wp/plugin/lib/cache.ml
fortunac: We currently have a script that runs wp on every function in a binary with a timeout. This created a large overhead trying to load the binary each time, so we tried saving the program_t and Arch.t into the knowledge base cache. Saving the entire KB to the cache still used up a lot of RAM and time, so we experimented with only saving the program_t and Arch.t while pruning away the other information.

Alternatively, we can try moving the function loop into wp itself, but it seems that we are struggling with getting timeouts working in Ocaml.

Do you have any suggestions?

wp/plugin/lib/utils.ml
```diff
@@ -54,26 +54,29 @@ let clear_mapper : Term.mapper = object
...
-    prog
+    prog, Option.value_exn (Cache.Arch.load filepath)
   | None ->
     (* The program_t is not in the cache. Disassemble the binary. *)
     info "Saving program %s (%a) to cache.%!"
       filepath Data.Cache.Digest.pp knowledge_digest;
     let project = create_proj None loader filepath in
     let prog = project |> Project.program |> clear_mapper#run in
+    let arch = Project.arch project in
+    let () = Toplevel.reset () in
```
ivg, line 75: This will break the inner consistency of the knowledge base. You have created a project, which is a value that depends on the knowledge base state, then you reset the state and store only the project, therefore all the objects that the projet references disappear. Even worse, any consequent computations will create new objects without knowing that they were already created in the previous knowledge base and thus unwated aliasing will be created. In other words, it is totally undefined behavior. 

If you just want to save the `Program.t` what stops you from storing it directly? Why put it in the knowledge base?

wp/plugin/lib/cache.ml
ivg: it, will, but I believe that `draper` or `cbat` should be enough. I hope that you will be able to figure manage names in your ogranization.

wp/plugin/lib/cache.ml
ivg: Something is wrong, 
```
        Command being timed: "bap /bin/ls"
        User time (seconds): 2.34
        System time (seconds): 0.41
```
and
```
$ bap cache --info
Capacity:      4096 MB
Current size:    22 MB
```
So storing the full knowledge base of `/bin/ls` takes 22Mb of disk space and the loading time is 2 seconds (less if no objdump or radare2 involved). Can you confirm that you have roughly the same readings, i.e., do

````
bap cache --clean
bap /bin/ls  # warm up the cache
/usr/bin/time -v bap /bin/ls
bap cache --info
```

philzook58: Yes. I agree with these numbers
```
real	0m0.256s
user	0m0.060s
sys	0m0.173s
```

philzook58: In our experience, our tar binaries take 10mins to load the first time. 50s to load the populated complete knowledge base and 5s to load this pruned knowledge base or a saved pruned Program.t

wp/plugin/lib/utils.ml
fortunac: This is true. I can definitely create a cache for just the `Program.t`. This way, I wouldn't have to mess with the inner consistency of the KB. Thanks!

wp/plugin/lib/cache.ml
philzook58: Relevant to this discussion:

- https://github.com/draperlaboratory/cbat_tools/issues/245 - An issue discussing how to parrallelize within the WP codebase.
- https://github.com/draperlaboratory/cbat_tools/issues/249 Our current method for running batches
- https://gitter.im/BinaryAnalysisPlatform/bap?at=5f0626978342f4627401cebb A discussion a bit back discussing this using the old cbat save-project plugin

philzook58:
```
philip@philip-desktop:$ time bap ../binaries/tar_amd64_linux_gcc_o0_dynamic

real	4m45.780s
user	4m43.940s
sys	0m1.678s
philip@philip-desktop:$ time bap ../binaries/tar_amd64_linux_gcc_o0_dynamic

real	0m8.525s
user	0m7.922s
sys	0m0.528s
```
These numbers are multiplied by at least 2 since we load 2 binaries.

ivg: What about picking the following strategy. First, disable cache on all stages, i.e., run bap with `--no-cache` option. Second, store only the pruned `program` along with any information that you need (e.g., arch, etc) in a file (not in cache). This will work if you don't perform any computations that touch the knowledge base after you load the program. For example, you shall not create new terms (this will break the tid consistency). If you need this, then I suggest storing the knowledge base and loading it. The numbers do not look that much, I mean 8 seconds to load the project is not that tragic. Let's continue in #245, as I think it is more about parallelizing, rather than about this PR.

wp/plugin/lib/utils.ml
ivg: I would even suggest just storing in a file without using cache. So that you can disable cache and work only with Program.t or even with Sub.t to enable the granularity that you need.

fortunac: I have two ideas:
1. Remove our use of the knowledge and state caches. Replace it with caches for `Program.t` and `Arch.t`. If its possible, combine those into the same cache.
2. We bring back our `save_project` plugin, but we serialize both `Program.t` and `Arch.t` information to our `.bpj` file. We create a flag that indicates that the file we are reading is a `.bpj` file rather than the binary itself.

I'm more inclined to the first idea.

ccasin: I also have a mild preference for the first idea, primarily because it keeps the CLI clean.

ivg: You can also specify your own loader that will read your stored data into the `Projec.t` representation, in #249 I even suggested to store one subroutine per file to minimize the bap startup overhead so that parallelizing wp could be done easily.

PR #250, opened 2020-09-25
Create a program cache

fortunac:
Based off the discussion from #248.

That PR broke the inner consistency of the knowledge base, something we should avoid. There were multiple solutions such as creating a custom loader to read our custom Project.t representation, saving one subroutine per file, or creating a cache for the Program.t

While having one subroutine per file sounds like a good idea to minimize the BAP startup overhead and the weakest precondition is only being calculated on one subroutine at a time, WP still looks at the other functions to create function summaries. We also want to avoid serializing our project to keep our CLI cleaner. Creating a program cache also required the least amount of changes to WP.

The speed of this is similar to the previous PR (e.g. `tar_dirname`: 8.645s and `parse_branch`: 2m46.430s.)

### DIFF
--- a/wp/plugin/lib/cache.ml
+++ b/wp/plugin/lib/cache.ml
@@ -16,7 +16,6 @@ open Bap_main
 open Bap_core_theory
 open Bap.Std
 open Regular.Std
-open KB.Let_syntax
 
 module Digests = struct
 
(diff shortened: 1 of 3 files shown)


wp/plugin/lib/cache.ml
```diff
@@ -39,6 +38,10 @@ module Digests = struct
   let project (mk_digest : namespace:string -> digest) : digest =
     mk_digest ~namespace:"project"
 
+  (* Creates a digest for the program state cache. *)
+  let program (mk_digest : namespace:string -> digest) : digest =
+    mk_digest ~namespace:"program"
```
philzook58, line 43: Is this at all related to Ivan's suggestion of namespacing or is this a separate thing?

fortunac: I removed the package name since we aren't storing the program in the knowledge base anymore. This is used to create the digest for the program cache, which is separate.

philzook58:
Looks good! I hope this puts us in compliance with proper use of the knowledge base.

Issue #237, opened 2020-09-01
hash_function multiple solutions


DieracDelta:
The `hash_function` test has too many solutions to feasibly iterate through. An idea to get around this is to, for each test, provide a function that checks that the registers represent a correct solution and a list of registers to search for with regex.

Consider `check_result`. We can iterate through the provided stream and generate a map of observed registers to values. Then to check the result, for each test we can provide to `check_result` a function of the form:

`is_valid_solution: ( register_mapping : string Map.t ) -> bool`

that determines whether or not the test's solution is valid. 

This will also make it easier to check solutions in sudoku and nqueens, since rather than enumerating through all possible solutions, we can simply check the solution.

DieracDelta:
closed by #253

PR #253, opened 2020-10-05
integration tests

DieracDelta:
This PR addresses #237, and adds in a couple of changes:

- The use of functions to check if test output is correct for plugin tests. These have been added for:
  - nqueens
  - 2 by 2 sudoku
  - hash function
- Colored output. This makes the test output a bit easier for me to read. Do yall think this is useful? If not, I can easily revert.
- The plugin tests have been split out into unit tests and integration tests in the same style as the unit and performance tests in the wp library. Additional makefile targets have been added; integration tests are not run by default (`make test`) as they take a while and include a couple timeouts.
    - To run these integration tests, `make test.plugin.integration` from the `wp` directory or `make test.integration` from the `plugin` directory.
- Two test changes have been added:
  - Multiple nqueens tests ranging from 4 to 19 inclusive. This adds in a
    dependency on clang. However, this should be easy to install (`apt-get
    install clang`). Do we have a place to list dependencies?
  - A 3 by 3 sudoku solver implementation. Unfortunately, the constraint encoding is too slow (and will time out). This is not surprising, as it encodes many more constraints than n-queens. I've got it as skipped right now, but maybe it makes sense to move this to the private repo instead.


wp/lib/bap_wp/Makefile
```diff
@@ -68,11 +68,11 @@ test: test.unit test.performance
 
 .PHONY: test.unit
 test.unit:
-	dune runtest tests/unit
+	OUINT_CI=true dune runtest tests/unit
```
fortunac, line 71: What does this env var do? Should it be `OUNIT_CI`?

wp/plugin/tests/test_libs/test_wp.ml
```diff
@@ -0,0 +1,577 @@
...
+
+let nqueens_tests = List.range 4 20 |> List.map ~f:(fun i ->
+    let description = Printf.sprintf "NQueens solver %dx%d" i i in
+    let script = Printf.sprintf "run_wp.sh" in
+    (description >: test_plugin "nqueens" sat
+       ~reg_list:(get_register_args 6 `x86_64 |> StringSet.of_list)
+       ~checker:(check_n_queens i `x86_64 |> Some)
+       ~script:script ~args:[string_of_int i]))
+
+let integration_tests = List.append [
```
fortunac, line 550: This file is getting rather large. What are your thoughts on separating the unit tests and integration tests into two different files?

wp/plugin/tests/test_libs/test_wp.ml
```diff
@@ -0,0 +1,577 @@
...
+(*  Government and no official endorsement should be inferred.             *)
+(*                                                                         *)
+(***************************************************************************)
+
+open !Core_kernel
+open Bap.Std
+open OUnit2
+open OUnitTest
+
+module StringMap = Map.Make(String);;
```
fortunac, line 19: This is a little nit-pick, but why do we need the `;;` over here?

codyroux: Don't use `;;` in `.ml` files, as a rule.

fortunac:
Tests look good! I'm currently getting three timeouts on the integration tests, on nqueens 17, 18, and 19. Is this expected? If not, what are your thoughts an allowing the use to specify their own timeout when invoking the tests? We can leave that for a separate PR since it's a separate issue.

wp/plugin/tests/test_libs/test_wp.ml
DieracDelta: Good point! Seems to compile fine without `;;`, so I'll remove them.

DieracDelta:
Based on @fortunac's feedback, I've made some changes:

- Removed unnecessary semicolons
- Skipped n=17-19 in nqueens, since these test cases time out right now.
- Split `test_wp.ml` into several files:
  - `test_wp_utils.ml` which contains helper functions used by both the unit and integration tests.
  - `test_wp_unit.ml` which declares the unit test suite.
  - `test_wp_integration.ml` which declares the integration test suite.

docs/installation.html
```diff
@@ -72,12 +72,24 @@ <h3>Installing CBAT</h3>
...
 </pre>
 
-<p>To run oUnit tests:
+<p>To run oUnit unit tests:
 
 <pre>
   make test
 </pre>
 
+<p>To run oUnit integration tests:
```
fortunac, line 81: Can we add this to the site?

fortunac (commented):
Looks good! After ~~adding that target to the site and~~ rebasing, I have no further changes requested. 🎉

PR #258, opened 2020-10-27
Fortunac/address rewrite

fortunac:
This PR adds the `--rewrite-addresses` flag which maps addresses of symbols in the modified binary to their addresses in the original binary.

For example, if a symbol `foo` was at address `0xdeadbeef` in the original and `0xcafef00d` in the modified, every instance of `0xcafef00d` found in the modified subroutine would be replaced with `0xdeadbeef`.


wp/resources/sample_binaries/memory_samples/diff_data_location/run_wp_addr_rewrite.sh
```diff
@@ -0,0 +1,18 @@
+# Tests having different locations for the data section and same values on the
+# stack. The binaries are the same except for the location of val.
+
+# This test turns on the rewritee-addresses flag, which rewrites the addresses
```
gltrost, line 4: *rewrite

wp/plugin/tests/unit/test_wp_unit.ml
```diff
@@ -43,20 +43,28 @@ let unit_tests = [
     ~script:"run_wp_pre.sh";
   "Arrays in data section"         >: test_plugin "memory_samples/arrays" unsat
     ~script:"run_wp_pre_mem_offset.sh";
+  "Arrays in data section"         >: test_plugin "memory_samples/arrays" unsat
+    ~script:"run_wp_addr_rewrite.sh";
 
   "Data/BSS sections"              >: test_plugin "memory_samples/data_bss_sections" sat;
   "Data/BSS sections"              >: test_plugin "memory_samples/data_bss_sections" unsat
     ~script:"run_wp_mem_offset.sh";
+  "Data/BSS sections"              >: test_plugin "memory_samples/data_bss_sections" unsat
```
gltrost, line 52: FWIW, running this case with `make test` gave me a `Timeout after 20.0s` but running `./run_wp_addr_rewrite.sh` by itself produced the correct answer. This is probably a bug on my end, but if a third person could run the tests on this branch and make extra sure this passes, that would be helpful.

fortunac: Ooh, maybe 20 seconds is too short of a timeout since we are on different machines. Did running the script by itself also take a while to run?

gltrost: Strangely it only takes 4 seconds individually. I'm trying it on another machine and all the tests pass now. There's no huge real to change the time limit if it's just my system acting up.

fortunac: I'm guessing it takes a while since we spin up a process that runs `objdump` when that flag is set, and IIRC the tests run in parallel. If that's the case, I think there is no problem bumping up the timeout.

gltrost (approved):
I see no issues.

wp/plugin/tests/unit/test_wp_unit.ml
fortunac: I changed the timeout for this test to `Short` = 60 seconds. Does this work better?

Issue #254, opened 2020-10-14
Abstract Equivalent Blocks


philzook58:
A hypothesis is that maybe z3 is getting stuck in the weeds because it can see the details of these blocks, when for comparative purposes they are irrelevant. Large sections of a program may be completely equivalent and only small differences occur.
In this case, it may be useful to abstract blocks to a set of uninterpreted functions, one for each variable that is written into in the 
block. Then the only thing z3 can see about these blocks is that they are the same between comparisons.

It is possible to separate the straight line computation computation from the control flow that occurs at the end of the block. We'd leave behind the jumps and only abstract the straight line computation

Consider the following two blocks
```
blockA
rsp := rax + 7
rax := rsp * 2

blockB
rax := rsp + 7
rsp := rax * 2
```

They can be seen to be equivalent by a "unification" of their variables (it's a little easier than the usual unification since we can consider their variable sets to be disjoint).

Here both would could be abstracted by
```
blockA
simultaneous{
rsp := f_1(rsp,rax)
rax := f_2(rsp,rax)
}

blockB
simultaneous{
rsp := f_1(rax,rsp)
rax := f_2(rax,rsp)
}
```

This hypothesis may not be true.  In this example, the abstracted version seems more complicated than the original. Maybe it's only worth doing for large blocks?

If Z3 returns unsat, great. That is a real unsat. If z3 returns SAT, that may be an erroneous SAT. One can then either just try the unabstracted version, or trace through the program to find the blocks touched and unabstract one or all of the blocks on that path and try again.

The simple version is to just include every variable that occurs in the right hand side in the block as an argument to the "block functions", but with a data flow analysis, it could be compressed to
```
blockA
simultaneous{
rsp := f_1(rax)
rax := f_2(rax)
}
```
Since we can see that original value of rsp does not affect anything

"simultaneous" could be achieved by using temporary variables, or by using Subst of Constr.t, which supports simultaenous substitution

Question: how much of a problem is memory? Memory is written to as a complete overwrite in bap. We would not want to do this in the abstraction, since then the entirety of memory could be scrambled by the block, when surely only a few locations are written to. It seems plausible through some analysis to only use the indexed into memory locations and only write to them though. Looking at bil, writes to memory are almost always in the shape of a simple store from a single location. If such an analysis fails and one knows it failed, one can always revert back to the unabstracted form of the block, no harm done.

PR #259, opened 2020-10-28
Short circuit when functions are syntactically equal

### LINKED ISSUE #211: Special Case Syntactically Equal Functions
During the bulk processing of two different binaries, some functions are not touched by the binary transformation. It would speed things up to special case easily detectable complete function equality.

DieracDelta:
This PR adds in a check for syntactic equality (enabled with `--syntax-equality` ). Example usage may be seen in `run_wp_*.sh` scripts in the `syntax_equality` test folder. This PR closes #211, and is a first step towards the ideas in #254. Specifically, 
`equality.ml` exports an `exist_isomorphism` function. This function checks if there's a mapping between the two `sub.t` functions that preserves syntax structure, use of virtual variables (virtual variables may go by different names but must be used in all the same places), and control flow.

I've documented the code in `equality.ml` and `equality.mli`. I'm also including the results I posted to the slack for the run on grammatech. On average, only running the syntax check takes 8 seconds on my laptop. I found 65 syntactically equal functions on grammatech. When running CBAT, I got:
- On the master branch cbat without any special flags: 36 SATS, 3 UNKNOWN, 26 UNSAT
- On the master branch cbat with mem-offset and pointer flag on all registers: 0 SAT, 7 UNKNOWN, and 58 UNSAT

So overall, the results that CBAT finishes within the timeout of 1000s turn out to match with this syntax check. It's not clear to me how to test the UNKNOWNs further. 

This first plot is function (on the x axis) vs # syntactically idental blocks / total blocks in the sub. These are encouraging numbers performance improvements in #254 .
image

This second plot is time in seconds on the x axis vs # statements (def_t, jmp_t) on the y axis. The green dots are the run for syntactically equal functions (of which there are 65); the red dots are master branch cbat without any special flags. @philzook58  mentioned that this is big savings if there are a lot of long functions that are syntactically identical in the binary.  
image


philzook58:
We may want equality.ml in `bap_wp` not `plugin`.

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
+open !Core_kernel
+open Bap.Std
+open Graphlib.Std
+open Base__Option.Monad_infix
+
+(* some of these pattern matchings are massive; we don't want fragile warnings *)
+[@@@warning "-4"]
+
+module H = Bap.Std.Graphs.Ir
```
philzook58, line 9: Is there a logic to these names? What is H? Maybe `Ir` would be a better choice if it doesn't slash with anything

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
...
+module J = Graphlib.To_ocamlgraph (H)
+module BFS = Graph.Traverse.Bfs (J)
+module TidSet = Tid.Set
+
+module TidMap = Tid.Map
```
philzook58, line 14: Stylistically, I don't think you're saving much with these shorthands.

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
...
+module VarMap = Var.Map
+module VarSet = Var.Set
+module TidTupleMap = Map.Make(struct type t = Tid.t * Tid.t [@@deriving sexp, compare] end)
+
+type varToVarMap = Var.t VarMap.t
+
+(* Replaces the real variables with indices with only their base *)
+let rec strip_indices (e : Exp.t) : Exp.t =
```
philzook58, line 22: I think this could be expressed more succinctly using the `mapper` interface. I think this is mapping Var.base over Var constructors and that's it?

wp/plugin/lib/equality.ml
DieracDelta: hmm not really. I'll swap it out for something a bit more descriptive (`Ir` if I can).

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
...
+    (Term.enum phi_t blk2 |> Sequence.to_list) map
+
+(* Iterate through all reachable nodes in each cfg
+ *  and generate a map from tid to blk.*)
+let get_node_maps
+    (cfg1 : Bap.Std.Graphs.Ir.t)
+    (cfg2 : Bap.Std.Graphs.Ir.t) : ((blk term) TidMap.t) * ((blk term) TidMap.t) =
+  let acc = fun ele tid_map ->
+    let tid = Term.tid (get_label ele) in
+    TidMap.change tid_map tid ~f:(fun _ -> Some (get_label ele)) in
```
philzook58, line 216: http://binaryanalysisplatform.github.io/bap/api/odoc/bap/Bap/Std/Tid/Map/index.html#val-update TidMap.update seems more appropriate

wp/plugin/lib/equality.mli
```diff
@@ -0,0 +1,30 @@
...
+     and
+        #88 := RBP
+        mem := mem with [RSP, el]:u64 <- #90
+     then the blocks would not be syntactically equal since #45 would be expected
+     in the spot that #90 is taking.
+   - The block TIDs mentioned within jmp_ts in sub1 when run through the
+     constructed block mapping must match the corresponding TIDS
+     in the same respective jmp_t in sub2.
+*)
+val exist_isomorphism : Sub.t -> Sub.t -> bool
```
philzook58, line 30: Something to consider, I can see it either way. Since you find an isomorphism, it seems it might be useful to expose it via returning `iso option` rather than `bool`

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
...
+  compare_defs (Term.enum def_t blk1 |> Sequence.to_list)
+    (Term.enum def_t blk2 |> Sequence.to_list) map >>=
+  fun map -> compare_jmps (Term.enum jmp_t blk1 |> Sequence.to_list)
+    (Term.enum jmp_t blk2 |> Sequence.to_list) map >>=
+  fun map -> compare_phis (Term.enum phi_t blk1 |> Sequence.to_list)
+    (Term.enum phi_t blk2 |> Sequence.to_list) map
+
+(* Iterate through all reachable nodes in each cfg
+ *  and generate a map from tid to blk.*)
+let get_node_maps
```
philzook58, line 211: Since the two cfgs and maps are built entirely independently, couldn't one make a `get_node_map : Ir.t -> blk TidMap.t` function that you call twice instead?

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
...
+module J = Graphlib.To_ocamlgraph (H)
+module BFS = Graph.Traverse.Bfs (J)
+module TidSet = Tid.Set
+
+module TidMap = Tid.Map
+module VarMap = Var.Map
+module VarSet = Var.Set
+module TidTupleMap = Map.Make(struct type t = Tid.t * Tid.t [@@deriving sexp, compare] end)
+
+type varToVarMap = Var.t VarMap.t
```
philzook58, line 19: I would prefer you not define this type synonym. `Var.t VarMap.t` is clearer and more self explanatory than `varToVarMap`

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
...
+  let tid2_map =
+    BFS.fold acc TidMap.empty cfg2 in
+  tid1_map, tid2_map
+
+(* Compares a sub to another sub; returns a map from tid into sub1
+   to the set of sub2 tids that it is syntactically equal to. Also returns
+   a map from (tid_blk_sub1, tid_blk_sub2) to the mapping between variables
+   used in those subs. *)
+let compare_blocks (sub1: Sub.t) (sub2 : Sub.t) :
+  (TidSet.t TidMap.t) * (varToVarMap TidTupleMap.t) =
```
philzook58, line 228: I have a suspicion that it would be better to use `((Var.t Var.Map.t)  Tid.Map.t) Tid.Map.t` rather than `varToVarMap TidTupleMap.t`. Essentially, currying the tid keys in that dictionary

wp/plugin/lib/equality.ml
```diff
@@ -0,0 +1,387 @@
...
+
+  let cfg1, cfg2 = Sub.to_cfg sub1, Sub.to_cfg sub2 in
+  (* blk1 tid -> set{blk2 tids} *)
+  let blk_map = TidMap.empty in
+  (* blk1 tid, blk2 tid -> varmap *)
+  let blk_varmap = TidTupleMap.empty in
+  let evaluator = (fun (node1 : Bap.Std.Graphs.Ir.Node.t) (blk_map, blk_varmap) ->
+      let blk1 = get_label node1 in
+      let tid1 = Term.tid (get_label node1) in
```
philzook58, line 237: could be blk1 instead of repeating (get_label node1)

DieracDelta:
Thanks for the suggestions! I've made corresponding changes. 

I agree--returning the mapping allows for more flexibility. I think the currying suggestion is clever--it makes exposing functions with aliased types easier to include in the `mli` file. Specifically, this lets me also expose `compare_blocks_syntax`. This function returns the syntactically equal (up to TID) blocks. Seems useful for #254 .

This `mapper` interface is awesome. I'm wondering if there's an easy way to use it to rewrite `sub_exp` in a cleaner fashion that doesn't involve fragile pattern matching. The issue is we need to carry around the information from both subs at the same time, and return something different if the subs don't match. This creates differing type signatures for the declared functions.

Also, I'm making liberal use of `>>=` and `>>|`. I could use `let` syntax instead if that's clearer. Seems like the `let` syntax is used in bildb, but the inline operators are used in `wp`. I'm not sure what's preferred here.

I'm testing the changes by running them on grammatech. I'll post again when I've got results.

wp/plugin/lib/equality.ml
codyroux: You can, I think module names shadow in OCaml (sometimes you need to be careful).

wp/plugin/lib/equality.mli
codyroux: Also: why isn't `exists`? Or even better, `get_iso`?

wp/plugin/lib/equality.ml
codyroux: Fun stuff.

codyroux:
There's no real norm between using monadic lets and `>>=`s, though Ivan prefers the latter, and his opinion has some weight.

Apparently, using the ppx versions of let bindings is deprecated, instead using the (newish) feature that allows defining new `let`-like keywords.

I'd suggest just using the combinators for now.

wp/lib/bap_wp/src/equality.ml
```diff
@@ -0,0 +1,354 @@
...
+             (* this should never exist *)
+             Tid.Map.update blk_varmap tid1
+               ~f:(fun cur_map ->
+                   match cur_map with
+                   | None -> Tid.Map.singleton tid2 v_map
+                   | Some m -> Tid.Map.add_exn m ~key:tid2 ~data:v_map)
+           | None ->
+             blk_map, blk_varmap) in
+      let blk_map, blk_varmap = BFS.fold inner_evaluator (blk_map, blk_varmap) cfg2 in
+      blk_map, blk_varmap) in
```
philzook58, line 220: Put this tuple in parens

wp/lib/bap_wp/src/equality.ml
```diff
@@ -0,0 +1,354 @@
...
+    List.fold z ~init:(Some map)
+      ~f:(fun (map_wrap) (j1, j2) ->
+          match map_wrap with
+          | Some map -> compare_jmp j1 j2 map
+          | None -> None)
+
+(* Compares everything about a block EXCEPT tids, which will be done later. *)
+let compare_block (blk1 : blk term) (blk2 : blk term) map =
+  compare_defs (Term.enum def_t blk1 |> Sequence.to_list)
+    (Term.enum def_t blk2 |> Sequence.to_list) map >>=
```
philzook58, line 176: Do we need to convert to lists?

wp/lib/bap_wp/src/equality.ml
```diff
@@ -0,0 +1,354 @@
...
+      Some (Var.Map.add_exn vmap ~key:v1 ~data:v2, v2)
+      (* if is physical register, just return original variable *)
+    else Some (vmap, v1)
+  (* if found in map, return what is found *)
+  | Some v_found -> Some (vmap, v_found)
+
+(* Maps all virtual variables from e1 to their analagous variable within vmap;
+   if a variable not in vmap, is added to vmap from e2.
+   Returns None if cannot map variables because subs do not match in structure *)
+let rec sub_exp (vmap : Var.t Var.Map.t)
```
philzook58, line 39: I'd separate this out into two functions. One that performs substitution and one that compare_exp

wp/lib/bap_wp/src/equality.ml
```diff
@@ -0,0 +1,354 @@
...
+    | None -> None
+
+(* Gets all syntactically equal blocks between the two subs,
+   Checks if it is possible to construct a mapping between
+   the two sets of blocks that matches in control flow. *)
+let subs_to_isomorphism (sub1: Sub.t) (sub2 : Sub.t) : (Tid.t Tid.Map.t) option =
+  let cfg1, cfg2 = Sub.to_cfg sub1, Sub.to_cfg sub2 in
+  let tid_to_blk1, tid_to_blk2 = get_node_map cfg1, get_node_map cfg2 in
+  let tid_map, var_maps = compare_blocks_syntax sub1 sub2 in
+  let used_set = Tid.Set.empty in
```
philzook58, line 346: Move into get_isomorphism

wp/lib/bap_wp/src/equality.ml
```diff
@@ -0,0 +1,354 @@
...
+   - [merged_map], a map built out of block-wise maps corresponding
+     to the current partial mapping
+   - returns a isomorphism from tids in sub1 to tids in sub2 if they exist *)
+let rec get_isomorphism
+    (candidate_map : Tid.Set.t Tid.Map.t) (used_set : Tid.Set.t)
+    (node_stack : (blk term) list) (graph : Tid.t Tid.Map.t)
+    (tid_to_blk1 : (blk term) Tid.Map.t) (tid_to_blk2 : (blk term) Tid.Map.t)
+    (var_maps : ((Var.t Var.Map.t) Tid.Map.t) Tid.Map.t) (merged_map : Var.t Var.Map.t) :
+  (Tid.t Tid.Map.t) option  =
+  let node = List.hd node_stack in
```
philzook58, line 313: This is a fold over the node_list I think

wp/plugin/lib/equality.ml
codyroux: (also, `var` is an alias to `Var.t` if that helps)

PR #260, opened 2020-11-23
Fortunac/valid mem derefs

fortunac:
This PR adds the `--check-invalid-derefs` flag, which adds an assumption to each memory read/write in the original binary that the address is in a valid location (i.e. above RSP or below the bottom of the stack), and asserts this property at every read/write in the modified binary.


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1054,6 +1054,102 @@ let non_null_store_assert : Env.exp_cond = fun env exp ->
...
+  (* addr >= RSP *)
+  let uge = BV.mk_uge ctx addr stack_pointer in
+  (* addr <= stack_bottom - 0x256 *)
+  let ule = BV.mk_ule ctx addr sb_bv in
+  (* addr >= RSP \/ addr <= stack_bottom - 0x256 *)
+  Bool.mk_or ctx [uge; ule]
+
+(* At every memory load, add a constraint that the address is in a valid
+   location in memory. *)
+let collect_valid_mem_loads (env : Env.t) (exp : Exp.t) : Constr.z3_expr list =
```
gltrost, line 1077: I notice the code for `collect_valid_mem_loads` and `collect_valid_mem_stores` only differ by one line. 
Likewise, `valid_load_vc` and `valid_store_vc` are nearly identical, and so are `valid_store_assert` and `valid_load_assert`

Should these pieces of identical-code not be abstracted? I think it looks fine as you wrote it, but I wanted to hear your (@fortunac) opinion.

fortunac: I can abstract out the similarities in `valid_load_vc` and `valid_store_vc` by creating a separate function and passing in a `collector` and the `name`, so that is no problem. As for `collect_valid_mem_loads` and `collect_valid_mem_stores`, I'm unsure how to abstract out the similarities in objects where I am overriding a method. Do you have suggestions for that?

gltrost: > I'm unsure how to abstract out the similarities in objects where I am overriding a method. Do you have suggestions for that?

I don't, and I figured that may have been an issue. Come to think of it, maybe declaring a new method should be more explicit anyway, so abstracting this out isn't really all that better.

codyroux: This is a weirdly non-modular aspect of OCaml. I wish we had a list to ask these kinds of questions.

fortunac: Okay, I pushed the changes for `valid_load_vc`, `valid_store_vc`, `valid_load_assert`, and `valid_store_assert`. I left `collect_valid_mem_loads` and `collect_valid_mem_stores`the same.

wp/resources/sample_binaries/memory_samples/mem_hooks/main_2.S
```diff
@@ -0,0 +1,23 @@
+.global _start
+
+.text
+
+main:
+    mov     (val),       %rax
+    nop
```
gltrost, line 7: Is this `nop` necessary? It didn't change the test-result when I removed it.

fortunac: Nope, not necessary. I just had it as padding to make the addresses different, but adding the `foo` symbol in the modified binary was sufficient enough. I can take this out!

gltrost: Fine either way. Actually maybe leaving it in to remind readers " `nop`'s can change addresses" is a good idea.

gltrost (approved):
The code, including the style and comments, looks great. The test cases pass.

PR #270, opened 2020-12-15
Wrap results of jumps in conditionals

fortunac:
This PR handles conditional calls (of the form ` when x call y with return z`) using the same functionality as our conditional jumps.


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -728,51 +728,54 @@ let mk_env
...
+    (Expr.to_string cond_val) (hooks_to_string hooks);
+  let cond_size = BV.get_size (Expr.get_sort cond_val) in
+  let false_cond = Bool.mk_eq ctx cond_val (z3_expr_zero ctx cond_size) in
+  let is_unconditional =
+    match cond with
+    | Bil.Types.Int w -> Word.is_one w
+    | _ -> false
+  in
+  let ite =
+    if is_unconditional then
```
gltrost, line 746: In `conditional_call/run_wp.sh`, we test a situation when `ite` is assigned `Constr.mk_ite jmp (Bool.mk_not ctx false_cond) target_pre post`, since we create `if not(= (ZF0) 0) then ...`. However, if I understand this code correctly, we don't test the case when `ite` is assigned a value based on `is_unconditional`. Would it be worth adding a test case for this `is_unconditional` case?

wp/resources/sample_binaries/conditional_call/main.asm
```diff
@@ -0,0 +1,15 @@
+        global __assert_fail:function (__assert_fail.end - __assert_fail)
```
gltrost, line 1: Stylistically, I don't think these don't need to be indented. 
Otherwise, this is a great, simple test-case. Well done!

jtpaasch: Don't think they don't need to be indented? So they do need to be indented?

wp/lib/bap_wp/src/precondition.ml
codyroux: Generally, it's a pretty good idea to try to test all code paths. I don't have particularly strong feelings about this specific case though. We're already pretty delinquent in terms of coverage.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -728,51 +728,54 @@ let mk_env
   Env.mk_env ~subs ~specs ~default_spec ~indirect_spec ~jmp_spec ~int_spec ~exp_conds ~num_loop_unroll
     ~arch ~freshen_vars ~use_fun_input_regs ~stack_range ~data_section_range ctx var_gen
 
+let conditional_jmp (jmp : Jmp.t) (env : Env.t) (target_pre : Constr.t)
```
gltrost, line 731: In my opinion, it would a good idea to add a description of `conditional_jmp` to `precondition.mli` as well.
Any thoughts?

wp/lib/bap_wp/src/precondition.ml
fortunac: I can definitely add one!

wp/resources/sample_binaries/conditional_call/main.asm
gltrost: Edit: I don't think these need to be indented. Apologies.

wp/lib/bap_wp/src/precondition.ml
fortunac: I agree, I should document this function. There's nothing written about this right now.

fortunac: I added documentation for `conditional_jmp` in `precondition.ml`. I did not want to expose this function in the `mli`, since it's a helper for `visit_jmp` and shouldn't really be called on its own.

wp/lib/bap_wp/src/precondition.ml
gltrost: With `sample_binaries/unconditional_call` added, we now cover the two conditions of `ite` inside `conditional_jmp`! Thanks!

wp/lib/bap_wp/src/precondition.ml
gltrost: That sounds like a better idea.

gltrost (approved):
I think this code looks great and is sufficiently tested.
The test cases pass on my end as well.

PR #273, opened 2020-12-31
Fortunac/func name mapping

fortunac:
This creates a map of original function names to modified function names, with a new flag of the form `--func-name-map=<reg_orig>,<reg_mod>`. Multiple patterns can be used and are delimited with a `;`. For example, we can have `main,main;foo,test_foo`. By default, wp assumes that the functions have the same names. I struggled with finding an intuitive way for the user to input the regex, so any and all feedback is helpful.

Right now, the map lives in `analysis.ml`, but the plan is to move it into the `env` so that we can make use of it when we handle function calls.


gltrost (approved):
This code looks great, and the new tests added here all pass. I think the `;` delimiter is fine, though you may want a third opinion.

PR #274, opened 2020-12-31
Add Bir_to_bil.call to handle function-call creations in unit tests

gltrost:
Here, I've abstracted on subroutine-call handling in unit tests to better handle function-calls. This is done by creating the function `call` in `Bil_to_bir.ml`. Namely, `call` creates a function-call based on a BIL term and appropriate type. 

I also edited `.gitignore` to ignore `.DS_Store` files.

### DIFF
--- a/wp/lib/bap_wp/src/bil_to_bir.ml
+++ b/wp/lib/bap_wp/src/bil_to_bir.ml
@@ -137,3 +137,6 @@ let bil_to_sub (stmts : Bil.t) : sub term =
   let res_sub = Term.remove blk_t res_sub head_tid in
   let res_sub = Term.prepend blk_t res_sub hd_blk in
   res_sub
+
+let call (call_sub : sub term ) (arch : Bil.typ) : Bil.Types.stmt  =
+  Bil.(jmp (unknown (call_sub |> Term.tid |> Tid.to_string) arch))
--- a/wp/lib/bap_wp/src/bil_to_bir.ml
+++ b/wp/lib/bap_wp/src/bil_to_bir.ml
@@ -138,5 +138,5 @@ let bil_to_sub (stmts : Bil.t) : sub term =
   let res_sub = Term.prepend blk_t res_sub hd_blk in
   res_sub
 
-let call (call_sub : sub term ) (arch : Bil.typ) : Bil.Types.stmt  =
-  Bil.(jmp (unknown (call_sub |> Term.tid |> Tid.to_string) arch))
+let call (call_sub : sub term ) (typ : Bil.typ) : Bil.Types.stmt  =
+  Bil.(jmp (unknown (call_sub |> Term.tid |> Tid.to_string) typ))
--- a/wp/lib/bap_wp/src/bil_to_bir.mli
+++ b/wp/lib/bap_wp/src/bil_to_bir.mli
@@ -14,7 +14,7 @@
 (**
 
    This module provides utilities to translate from a BIL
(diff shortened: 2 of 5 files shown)


wp/lib/bap_wp/src/bil_to_bir.ml
```diff
@@ -137,3 +137,6 @@ let bil_to_sub (stmts : Bil.t) : sub term =
   let res_sub = Term.remove blk_t res_sub head_tid in
   let res_sub = Term.prepend blk_t res_sub hd_blk in
   res_sub
+
+let call (call_sub : sub term ) (arch : Bil.typ) : Bil.Types.stmt  =
```
codyroux, line 141: I'd call that argument `typ` or `type`. Obviously it does depend (somehow) on the architecture, but it's just the type of the call.

wp/lib/bap_wp/src/bil_to_bir.mli
```diff
@@ -49,4 +49,8 @@ val mk_assert_fail : unit -> Bap.Std.Sub.t * Bap.Std.Exp.t
 
 (** Take a BIL program (a list of statements) and turn it into a
     subroutine. *)
-val bil_to_sub : Bap.Std.Bil.t -> Bap.Std.Sub.t
\ No newline at end of file
+val bil_to_sub : Bap.Std.Bil.t -> Bap.Std.Sub.t
+
+(** Take a sub term and a type for the term to build a BIL
```
codyroux, line 54: I'd say [sub_t term] (with the brackets so it's pretty in the docs) or just "a subroutine".

wp/lib/bap_wp/src/bil_to_bir.ml
gltrost: Good idea. Changing the name to `typ` will also make it more consistent with the comments I've made.

wp/lib/bap_wp/src/bil_to_bir.mli
codyroux: More generally, I'd

A) `open Bap.Std` so it doesn't need to be prefixed everywhere

B) Use the "nice names" everywhere, so `sub term` instead of `Sub.t` etc.

I understand this is a bit of scope creep, since it doesn't directly pertain to the added functionality. Maybe @jtpaasch will tell me that's Not Ok.

codyroux:
This looks really nice! My only general comment would be some minor cleanup in the `mli`, but that's not even crucial.

wp/lib/bap_wp/src/bil_to_bir.mli
gltrost: > A) open Bap.Std so it doesn't need to be prefixed everywhere
>
> B) Use the "nice names" everywhere, so sub term instead of Sub.t etc.

It was looking pretty cluttered with the `Bap.Std`prefix and `t`s.

gltrost: @codyroux `B)` might cause style mismatching with `bil_to_bir.ml`. For example, I could change 
`mk_assert_fail : unit -> Sub.t * Exp.t` to `mk_assert_fail : unit -> sub term * Exp.t`, though in `bil_to_bir.ml` we have it written as `let mk_assert_fail () : Sub.t * Exp.t =`.

codyroux: It doesn't matter, the `.ml` is for our eyes only, but the `mli` is for any client of the module. Also, it would be `Bil.exp` rather than `Exp.t`.

But you can change the `.ml` also, if you like.

PR #277, opened 2021-01-18
Jt/update bildb

jtpaasch:
I've updated BILDB in two ways:
* Updated all the code for latest BAP and Core (so using new let bindings, and new Core methods and equalities).
* Bildb will now let you write to memory addresses that's not allocated (behind the scenes, bildb will first allocate that address, then write the word you want to write there).

### DIFF
--- a/bildb/lib/screens/locations.ml
+++ b/bildb/lib/screens/locations.ml
@@ -181,27 +181,15 @@ module Make (Machine : Primus.Machine.S) = struct
         | Some value' ->
           begin
 
-              (* Is the address we're going to write to unmapped?
-                 If not, we can write to it. *)
-              let* addrs = unmapped addr' ~addr_size ~num_words:1 in
-              match addrs with
-              | [] ->
-                begin      
-                  let* _ = Mem.store addr' value' in
-                  show_locs addr ~prompt ~handler ~num_words:"1"
-                end
-
-              (* If the address isn't mapped, alert the user. 
-                 Trying to write to it would result in a page fault,
-                 and cause Primus to exit. *)
-              | _ ->
-                begin
-                  let msg = Printf.sprintf 
-                    "The address is unmapped and can't be written to: %s" 
-                    (Utils.string_of addr') in
-                  let text = [Ui.mk_output ~color:Tty.Red msg] in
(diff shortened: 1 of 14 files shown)


bildb/lib/screens/locations.ml
```diff
@@ -180,27 +181,15 @@ module Make (Machine : Primus.Machine.S) = struct
...
-                begin
-                  let msg = Printf.sprintf 
-                    "The address is unmapped and can't be written to: %s" 
-                    (Utils.string_of addr') in
-                  let text = [Ui.mk_output ~color:Tty.Red msg] in
-                  Machine.return (Event.screen () ~text ~prompt ~handler)
-                end
+              (* Allocate the memory at the specified address (if needed),
+                 then store the provided value there. *)
+              let* is_mapped = Mem.is_mapped addr' in
```
gltrost, line 186: Could you give a brief description of the code being added here?... and/or more generally your use of `Machine.return` and `Mem.allocate`? 
(By brief description, I just mean in this pr, not actually adding comments in the code.)

jtpaasch: In general, all things we do in Primus are monadic actions. They have "side-effects" (I change values in memory, or in a register, or whatever). So they're all monadic types. Hence, any value we want to return must be returned with `Machine.return (value)`, which is the monadic return that the `Machine` module provides for us. 

Most of the time, when I make some change to the memory or registers, I don't need to return a value, so I finish that action with `Machine.return ()`.

`Memory.allocate` already has built into it that after it has allocated the memory, it will call `Machine.return ()`, so I don't have to actually call that. It will do it for me. 

In this particular place, both branches of this if-statement return the same type. On one branch, we do nothing (i.e., we return `Machine.return ()`). On the other branch, we allocate a bit of memory, and then do nothing (i.e., let `Memory.allocate` return `Machine.return ()`).

Does that help any?

gltrost: That helps a great deal! Thanks.

gltrost (approved):
The code looks good, and the test cases (except for `goto-string`) pass. I also grepped for `let%bind` and got nothing.

jtpaasch:
Thanks Greg!

PR #278, opened 2021-01-19
Fortunac/fun specs flag

fortunac:
Adds a flag so the user can specify which built-in function specs they want to use. The built-in summaries are:
- verifier-error
- verifier-assume
- verifier-nondet
- afl-maybe-log
- arg-terms
- chaos-caller-saved
- rax-out
- chaos-rax

Also checks which subroutines are to be inlined. If a subroutine is given both the inline spec and a summary spec, this will report an error rather than picking one spec over the other.

### DIFF
--- a/wp/plugin/wp.ml
+++ b/wp/plugin/wp.ml
@@ -221,31 +221,40 @@ let fun_specs = Cmd.parameter Typ.(list string) "fun-specs"
            site in order of precedence. A target function will be mapped to a
            function spec if it fulfills the spec's requirements. All function
            specs set the target function as called and update the stack
-           pointer. The default spec will set the function as called and update
-           the stack pointer. Available built-in specs:
+           pointer. The default specs set are verifier-assume, varifier-nondet,
+           chaos-caller-saved, and empty. Note that if a function is set to be
+           inlined, it will not use any of the following function specs.
+           Available built-in specs:
 
-           `verifier-error': Trips calls to __assert_fail.
+           `verifier-error': Used for calls to __VERIFIER_error and
+           __assert_fail. Looks for inputs that would cause execution to reach
+           these functions.
 
-           `verifier-assume': Adds an assumption to the precondition based on
-           the argument to the function call.
+           `verifier-assume': Used for calls to __VERIFIER_assume. Adds an
+           assumption to the precondition based on the argument to the function
+           call.
 
-           `verifier-nondet': Chaoses the output to the function call
-           representing an arbitrary pointer.
+           `verifier-nondet': Used for calls to nondeterministic functions such
+           as __VERIFIER_nondet_*, calloc, and malloc. Chaoses the output to
+           the function call representing an arbitrary pointer.
 
-           `afl-maybe-log': Chaoses the registers RAX, RCX, and RDX. Used for
-           calls to __afl_maybe_log.
+           `afl-maybe-log': Used for calls to __afl_maybe_log. Chaoses the
+           registers RAX, RCX, and RDX.
 
-           `arg-terms': Uses BAP's uplifter to determine the input and output
-           registers to a function and chaoses the output registers.
+           `arg-terms': Used when BAP's uplifter returns a nonempty list of
+           input and output registers for the target function. Chaoses this
+           list of output registers.
 
-           `chaos-caller-saved': Chaoses the called-saved registers for x86
-           architectures.
+           `chaos-caller-saved': Used for the x86 architecture. Chaoses the
+           caller-saved registers.
 
            `rax-out': Chaos RAX if it can be found on the left-hand side of an
            assignment in the target function.
 
            `chaos-rax': Chaos RAX regardless if it has been used on the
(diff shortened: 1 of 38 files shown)


wp/plugin/lib/analysis.ml
```diff
@@ -124,20 +124,10 @@ let exp_conds_mod (p : Params.t) : Env.exp_cond list =
 (* Determine which function specs to use in WP. *)
 let fun_specs (p : Params.t) (to_inline : Sub.t Seq.t)
```
codyroux, line 125: Per our Wednesday conversation: what is the default behavior if the user does not set the flag?

wp/resources/sample_binaries/nested_ifs/run_wp_inline.sh
```diff
@@ -8,7 +8,7 @@
 run () {
   bap wp \
     --func=main \
-    --inline=.* \
+    --inline=calloc \
```
codyroux, line 11: Is this a fix for an older issue?

wp/plugin/wp.ml
```diff
@@ -207,6 +207,36 @@ let func_name_map = Cmd.parameter Typ.(list ~sep:';' (pair ~sep:',' string strin
            "<reg1_orig>,<reg1_mod>;<reg2_orig>,<reg2_mod>"). By default, WP
            assumes subroutines have the same names between the two binaries.|}
 
+let fun_specs = Cmd.parameter Typ.(list string) "fun-specs"
+    ~doc:{|List of built-in function summaries to be used at a function call
+           site in order of precedence. A target function will be mapped to a
+           function spec if it fulfills the spec's requirements. All function
```
codyroux, line 213: Should we say more about the "spec's requirements"?

wp/plugin/wp.ml
```diff
@@ -207,6 +207,36 @@ let func_name_map = Cmd.parameter Typ.(list ~sep:';' (pair ~sep:',' string strin
...
+           specs set the target function as called and update the stack
+           pointer. The default spec will set the function as called and update
+           the stack pointer. Available built-in specs:
+
+           `verifier-error': Trips calls to __assert_fail.
```
codyroux, line 218: What is "Trips calls"?

wp/plugin/wp.ml
```diff
@@ -207,6 +207,36 @@ let func_name_map = Cmd.parameter Typ.(list ~sep:';' (pair ~sep:',' string strin
...
+
+           `verifier-assume': Adds an assumption to the precondition based on
+           the argument to the function call.
+
+           `verifier-nondet': Chaoses the output to the function call
```
codyroux, line 223: It feels like these correspond to specific functions.

wp/plugin/lib/analysis.ml
fortunac: The current default behavior is to set the function as called and update the stack pointer. I plan on changing this to chaosing the caller-saved registers, and if a function is one of the special functions from our verifier api, use those corresponding specs.

wp/resources/sample_binaries/nested_ifs/run_wp_inline.sh
fortunac: Since I removed VERIFIER_nondet from the default list of function specs, inlining everything, would inline calloc, and that would make this test fail. When I add the nondet spec back to the default, I think I could change this back to `.*`.

wp/plugin/wp.ml
fortunac: Can do!

wp/plugin/wp.ml
fortunac: It would get WP to look for inputs to the subroutine that would cause us to reach `__assert_fail' or `__VERIFIER_error'. I can add this.

wp/plugin/wp.ml
fortunac: I'll add this too.

wp/plugin/lib/analysis.ml
codyroux: 👍 just make sure this appears in the comments/docs somewhere.

wp/resources/sample_binaries/nested_ifs/run_wp_inline.sh
codyroux: 👍

wp/plugin/wp.ml
codyroux: I just want the docs to explain what "trips calls" mean to a casual user.

wp/plugin/wp.ml
codyroux: 👍

fortunac:
Changed the default specs to be `verifier-assume`, `verifier-nondet`, `chaos-caller-saved`, and `empty`. I did not add `verifier-error` to the default list of specs since it is controlled by the `trip-asserts` flag (although the user can stilll specify it with the `fun-specs` flag). I also updated the docs to include the list of default specs and the functions each spec corresponds to.

PR #281, opened 2021-01-25
Fortunac/package tests

fortunac:
This PR adds the binaries from our test suite to the repo. It updates the directory structure of each test case to separate the source code and binaries from each other:
- test_case/
  - bin/
    - main_1
    - main_2
  - src/
    - main_1.c
    - main_2.c
  - Makefile
  - run_wp.sh

Also stops building the test binaries every time we run `make test`. The target to build and clean all test binaries still exists in the `plugin` and `sample_binaries` directories.


codyroux:
How did you generate this patch?

codyroux:
Seems to work.

Issue #286, opened 2021-02-02
Make sure WP is ocp-indent compliant


fortunac:
We should run `ocp-indent` on all of WP to make sure the code's spacing is compliant.

gltrost:
I can quickly run on `ocp-indent -i` on all `.ml` and `.mli` files.

codyroux:
Doesn't seem to be a priority for now.

PR #287, opened 2021-02-02
Run ocp-indent on all .ml and .mli files in cbat_tools

gltrost:
Adresses issue #286. I ran `ocp-indent -i` on all `.ml` and `.mli` files I could manually find, including those outside of `wp`. This will allow our code to be more compliant with standard OCaml style.


codyroux:
Sorry for your hard work Greg, but it sounds like the consensus is that the bang/buck ratio is too low on this one.

PR #288, opened 2021-02-04
Add comparative tests to sample_binaries/user_func_specs

gltrost:
Currently, the flag `--user-func-specs` is only tested in the single-program case, but not in the comparative case. This branch adds two tests, `sub_spec_3/run_wp_comp.sh` and `sub_spec_1/run_wp_1.sh` to the comparative test, where adding the `--user-func-specs` flag changes a `SAT` result to an `UNSAT` result in both new cases.


wp/plugin/tests/unit/test_wp_unit.ml
```diff
@@ -241,9 +241,13 @@ let unit_tests = [
...
   "User defined sub specs 7" >: test_plugin "user_func_spec/sub_spec_3"
-    unsat ~script:"run_wp_1.sh";
+    unsat ~script:"run_wp_single_1.sh";
   "User defined sub specs 8" >: test_plugin "user_func_spec/sub_spec_3"
-    sat ~script:"run_wp_2.sh";
+    sat ~script:"run_wp_single_2.sh";
   "User defined sub specs 9" >: test_plugin "user_func_spec/sub_spec_3"
-    unsat ~script:"run_wp_3.sh";  
+    unsat ~script:"run_wp_single_3.sh";
+  "User defined sub specs 10" >: test_plugin "user_func_spec/sub_spec_3"
```
fortunac, line 249: Small nit. The list of tests are divided with comparisons on the top and single cases on the bottom, so I think it would be good to be consistent.

gltrost: Ok. I'll move those to their proper locations.

fortunac (approved):
Looks good to me! 🎉

PR #290, opened 2021-02-09
Fortunac/arm

### LINKED ISSUE #285: Support for ARM
Currently, WP makes assumptions that we are analyzing x86, most notably in the lists of registers we generate for function specs and assumptions about the stack pointer.
- https://github.com/draperlaboratory/cbat_tools/blob/47d7971d89ffd6119b1e77ba27b734972a12fe74/wp/lib/bap_wp/src/precondition.ml#L389 
- https://github.com/draperlaboratory/cbat_tools/blob/3621f5c1906403db8b382ac6c51ea95fbf2f16f4/wp/lib/bap_wp/src/precondition.ml#L624
- etc...
We should extend this to add support for ARM.

fortunac:
Fixes #285. Adds support for ARM, most notably:
- the function spec for chaosing caller-saved registers
- setting SP to the defined region
- printing the memory model when outputting gdb and bildb init scripts

I did not create ARM versions of the `chaos-rax` or `rax-out` specs in favor of using the `arg-terms` and `chaos-caller-saved-registers` specs. If we want to create a `chaos-R0` spec, that can be done easily.


wp/lib/bap_wp/src/output.ml
```diff
@@ -144,16 +144,11 @@ let format_model (model : Model.model) (env1 : Env.t) (env2 : Env.t) : string =
...
   let arch = Env.get_arch env in
-  if Env.is_x86 arch then
-    begin
-      let module Target = (val target_of_arch arch) in
-      let mem, _ = Env.get_var env Target.CPU.mem in
-      Some (extract_array (Constr.eval_model_exn m mem))
-    end
-  else
-    None
+  let module Target = (val target_of_arch arch) in
```
codyroux, line 149: Just as a note: we need to eventually fix this, so that we just grab the target from the KB.

A task for another ticket...

fortunac: I agree. There are lots of instances where we get the architecture from the environment to get the target.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -401,9 +401,16 @@ let input_regs (arch : Arch.t) : Var.t list =
...
+  | `x86 ->
+    warning "In 32-bit x86, arguments are passed through the stack.%!";
+    []
+  | #Arch.arm ->
+    let open ARM.CPU in
+    [r0; r1; r2; r3]
+  | a ->
+    let err = Format.sprintf "input_regs: Input registers have not been \
+                              implemented for %s." (Arch.to_string a) in
+    raise (Not_implemented err)
```
codyroux, line 413: I wonder if there's some more BAP idiomatic way of raising an error.

On another note, I wonder if we could just print a warning here and charge on regardless.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -412,9 +419,16 @@ let caller_saved_regs (arch : Arch.t) : Var.t list =
...
-    raise (Not_implemented "caller_saved_regs: Caller-saved registers have not been \
-                            implemented for non-x86_64 architectures.")
+  | `x86 ->
+    let open X86_cpu.IA32 in
+    [rax; rcx; rdx]
+  | #Arch.arm ->
+    let open ARM.CPU in
+    [r0; r1; r2; r3; r12]
+  | a ->
+    let err = Format.sprintf "caller_saved_regs: Caller-saved registers have not \
```
codyroux, line 429: Same comment as above: should we be violently failing?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -626,19 +654,16 @@ let spec_chaos_rax (sub : Sub.t) (arch : Arch.t) : Env.fun_spec option =
...
-          (fun env post tid ->
-             let post = set_fun_called post env tid in
-             let post, env = increment_stack_ptr post env in
-             let inputs = if Env.use_input_regs env then input_regs arch else [] in
-             let regs = caller_saved_regs arch in
-             subst_fun_outputs env sub post ~inputs ~outputs:regs)
-    }
-  else
-    None
+  Some {
```
codyroux, line 657: Great. This should prevent some confusion!

codyroux:
As usual, very clean and pleasant implementation!

We just need to agree on the default behavior on "non-supported" architectures: currently it fails with `--compare-regs`, I think. My intuition is that it should proceed with warning and an empty set?

Not sure.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -467,6 +488,13 @@ let rec get_vars (env : Env.t) (t : Sub.t) : Var.Set.t =
   in
   visitor#visit_sub t vars
 
+let get_vars (env : Env.t) (t : Sub.t) : Var.Set.t =
+  let gprs = Env.get_gprs env in
+  let mem = Var.Set.singleton (Env.get_mem env) in
+  let sp = Var.Set.singleton (Env.get_sp env) in
+  let sub_vars = vars_from_sub env t in
+  Var.Set.union_list [gprs; mem; sp; sub_vars]
```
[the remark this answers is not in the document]
fortunac: Didn't realize that function existed! It might be a good idea to return a union of the `sub_vars` and the `Target.vars` to capture everything including the virtual registers. I can test this out.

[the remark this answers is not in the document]
codyroux: Nah, that would be redundant. If we can use just `Target.vars` then we should, but I'm not sure it's available/doesn't break backwards compatibility.

Unioning both is ugly and will probably be problematic at some point.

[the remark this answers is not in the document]
fortunac: I tried testing it out. It breaks backwards compatibility since none of wp's library uses the core theory. Doing this would probably require a big rewrite.

fortunac:
> 
> 
> As usual, very clean and pleasant implementation!
> 
> We just need to agree on the default behavior on "non-supported" architectures: currently it fails with `--compare-regs`, I think. My intuition is that it should proceed with warning and an empty set?
> 
> Not sure.

I have no problems with printing a warning and returning an empty set.

wp/lib/bap_wp/src/precondition.ml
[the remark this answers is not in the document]
codyroux: Thought it might. Let's drop it for the time being.

Issue #291, opened 2021-02-10
Give Output.print_result a parameter to specify a formatter


jtpaasch:
Currently, in `wp`, `Output.print_result` prints everything to `stdout`. In fact, it uses `Printf.printf`. 

The purpose of this ticket is twofold:

(1) Let the user specify which channel they want to output the result to.
(2) The default should be stderr.

Concretely: add an optional parameter to `Output.print_result` which lets the user specify a formatter, which defaults to the `err_formatter` unless otherwise specified, and then use `fprintf` instead of `printf`. 

In the future, we should really be returning good exit codes that tell if we got sat or unsat, but for now, if we put our diagnostic information into stderr and allow the user to specify their formatter, we'll at least be a better *nix citizen.

jtpaasch:
Apologies for confusions on this. Here's a clarification: this ticket requires no changes to the CLI _plugin_. It only asks for changes to the _library_.

If the default formatter in the library is stderr, then the plugin behavior won't change. Everybody will still see their output printed to their tty just as they did before. 

However, this will give a _library_ user the power to specify a custom place for the output. Then the library user won't be forced to live with the output getting dumped directly to stdout.

codyroux:
You don't want the user to be able to set the output stream?

jtpaasch:
Not for the CLI. TTYs have redirection already. We don't need to reimplement that feature.

codyroux:
Fair enough.

PR #293, opened 2021-02-12
Address issue #291 by adding the formatter control

gltrost:
The following branch creates a new flag called `formatter` that allows the user to speficy if the results should be printed to `stderr` or `stdout`. If the flag is not set, the results default to `stderr`. 

This requires changes to `Output.print_result` and `Precondition.check`, which both print information. The function `Precondition.exclude`  was also altered to take into account the changes to `check` since `exclude` calls `check`.


jtpaasch (changes requested):
We don't need to change the CLI plugin. Giving the CLI user the option to print to stdout or stderr doesn't give them anything they don't already have with standard redirection techniques. 

So I would undo any of the changes in the `cbat_tools/wp/plugin` folder (e.g., undo the changes to `wp.ml`, and to parameters, and take out the bit that validates the `--formatter` param, etc.)

All we need is to give users of the _library_ the ability to specify where they want the output to go.

In the library functions, we should be able to let the user pass in a formatter directly, via a parameter `?fmt:(fmt = Format.stderr_formatter)`, and then we should be able to use `Format.fprintf fmt` to print things. E.g., I would alter `print_result` to be like this:

```
let print_result (solver : Solver.solver) (status : Solver.status) (goals: Constr.t)
    ?fmt:(fmt = Format.err_formatter)
    ~show:(show : string list) ~orig:(env1, sub1 : Env.t * Sub.t)
    ~modif:(env2, sub2 : Env.t * Sub.t) : unit =
  | Solver.UNSATISFIABLE -> Format.fprintf fmt "%s%!" "\nUNSAT!\n"
  | Solver.UNKNOWN -> Format.fprintf fmt "%s%!" "\nUNKNOWN!\n"
  | Solver.SATISFIABLE -> ...
```

And `check` would go like this:

```
let check ?fmt:(fmt = Format.err_formatter) ?refute:(refute = true) ?(print_constr = []) ?(debug = false)
    (solver : Solver.solver) (ctx : Z3.context) (pre : Constr.t)  : Solver.status =
  Format.fprintf fmt "Evaluating precondition.\n%!";
  if (List.mem print_constr "precond-internal" ~equal:(String.equal)) then (
    Format.fprintf fmt "Internal : %s \n %!" (Constr.to_string pre) ) ;
  let pre' = Constr.eval ~debug:debug pre ctx in
  Format.fprintf fmt "Checking precondition with Z3.\n%!";
  ... 
```

And so on. Wouldn't that work?

gltrost:
> And so on. Wouldn't that work?
@jtpaasch Yeah! That or something very close to that, should work. I'll make those changes now.

wp/lib/bap_wp/src/output.mli
```diff
@@ -31,9 +31,15 @@ module Constr = Constraint
     represents the registers and memory values that lead to a specific program state,
     a list of goals that have been refuted, and if specified, the paths that lead to
     the refuted goals. *)
-val print_result :
-  Z3.Solver.solver -> Z3.Solver.status -> Constr.t -> show:string list ->
-  orig:Env.t * Bap.Std.Sub.t -> modif:Env.t * Bap.Std.Sub.t -> formatter:(string option) -> unit
+val print_result
+    : ?fmt:Stdlib__format.formatter
```
jtpaasch, line 35: Can we write `Format.formatter` instead of `Stdlib__format.formatter`? I don't have any opinion on whether we should or should not do that. I'm just curious if there's a reason why you fully specified `Stdlib_format` instead of just `Format`. And by no means should this hold up merging this.

jtpaasch (approved):
Everything looks good to me. I made one comment, mostly out of curiosity, which you're welcome to look at or ignore. After the merge conflicts are fixed, I say merge it.

wp/lib/bap_wp/src/output.mli
gltrost: There's no deeper reason for using `Stdlib__format` instead of `Format`, and using `Format` would be more consistent. I'll change this now.

Issue #301, opened 2021-03-12
Failed to build `value_set`


whst:
The following line declared `equal` to be `t -> t -> bool`:
https://github.com/draperlaboratory/cbat_tools/blob/4ff5ae9bdfbfa85947f1bcb3e5141309fffb39ae/vsa/value_set/lib/src/cbat_lattice_intf.ml#L34

For the definition,
https://github.com/draperlaboratory/cbat_tools/blob/4ff5ae9bdfbfa85947f1bcb3e5141309fffb39ae/vsa/value_set/lib/src/cbat_lattice_intf.ml#L142
OCaml think the `(=)` is `Core_kernel.Int.t -> Core_kernel.Int.t -> bool` (even when we annotate it with `let equal (a:t) (b:t) = a = b`).

Changing the definition to `let equal a b = Bool.(a = b)` solves the problem here, but brings more similar typing errors.

----

Tested with 2 environments,
* Docker Ubuntu with `OCaml = 4.09.1` and  `bap = 2.3.0-alpha`
* Ubuntu with `OCaml = 4.08.1` and `bap = 2.2.0`

```
File "src/cbat_lattice_intf.ml", lines 133-147, characters 6-3:
133 | ......struct
134 | 
135 |   type t = bool[@@deriving bin_io, sexp, compare]
136 | 
137 |   let top = true
...
144 | 
145 |   let pp ppf (b : t) = Format.fprintf ppf (if b then "top" else "bottom")
146 | 
147 | end
Error: Signature mismatch:
       ...
       Values do not match:
         val equal : Core_kernel.Int.t -> Core_kernel.Int.t -> bool
       is not included in
         val equal : t -> t -> bool
       File "src/cbat_lattice_intf.ml", line 34, characters 2-28:
         Expected declaration
       File "src/cbat_lattice_intf.ml", line 142, characters 6-11:
         Actual declaration
Makefile:11: recipe for target 'build' failed
make[1]: *** [build] Error 1
make[1]: Leaving directory '/home/opam/cbat_tools/vsa/value_set/lib'
Makefile:10: recipe for target 'install' failed
make: *** [install] Error 2
```

ccasin:
Sorry!  We are aware the `value_set` plugin does not compile with BAP 2.x - it was built for BAP 1.  Our main work is on the `wp` plugin in this repo, these days, but we hope to find time to revisit the value set analysis soon.  I will update the readme to reflect the fact that `value_set` is known not to work.

[...]

Issue #298, opened 2021-02-25
value set build problem


zhouxuan009:
Hi, I entered `make` under the folder cbat_tools/vsa/value_set, then the make error message appeared in the command line as below
```
make -C lib install.local
make[1]: Entering directory '/open/zhouxuan/cbat_tools/vsa/value_set/lib'
dune build -p cbat_value_set
      ocamlc src/.cbat_value_set.objs/byte/cbat_value_set__Cbat_lattice_intf.{cmi,cmo,cmt} (exit 2)
[...]

PR #303, opened 2021-03-12
Update value-set analysis for BAP 2.x

whst:
I updated the VSA plugin for BAP 2.x (a somewhat workaround, though).

* Both `vsa/value_set` and `vsa/explicit_edge` now can be built
* All Unit tests in `value_set` passed
* Tested with `bap 2.2.0` and `OCaml 4.08.1`


ccasin (approved):
Looks good.  All changes related to picking the right equality.

Fixes #301 and #298

Issue #300, opened 2021-03-01
Improve readability our constraint printing


gltrost:
Currently when printing constraints, we print them in a hard to read syntax. For example: 
```
Internal : (SP in stack range: (and (bvult (bvadd (bvsub #x0000000040000000 #x0000000000800000)
                   #x0000000000000080)
            RSP0)
     (bvule RSP0 #x0000000040000000)): (let ((a!1 (not (and (= ((_ extract 63 30) RSP0)
                        #b0000000000000000000000000000000000)
                     (bvule ((_ extract 29 0) RSP0)
                            #b111111100000000000000010000000)))))
  (and a!1
       (= ((_ extract 63 31) RSP0) #b000000000000000000000000000000000)
       (bvule ((_ extract 30 0) RSP0) #b1000000000000000000000000000000)))|#540| = |init_\|#540\||: (= |#540| |init_\|#540\||)|#510| = |init_\|#510\||: (= |#510| |init_\|#510\||)|#480| = |init_\|#480\||: (= |#480| |init_\|#480\||)mem0 = init_mem0: (= mem0 init_mem0)ZF0 = init_ZF0: (= ZF0 init_ZF0)YMM90 = init_YMM90: (= YMM90 init_YMM90)YMM80 = init_YMM80: (= YMM80 init_YMM80)YMM70 = init_YMM70: (= YMM7
```
is difficult to understand. Things that should be fixed include

1. the explicitly long strings of zeros 
2. the `\n` printings
3. left-sided binary operations like `=`
4. long indentations 

If anyone has other requests with respect to printing constraints, please share!

codyroux:
I was actually hoping that `Constraint.to_string` could be significantly improved as well, e.g. by printing the constructors in an easy to read format, e.g. `ITE (a, b, c)` as `if a then b else c`, `Subst(a, b, c)` as `let b = c in a` etc.

codyroux:
Based on comments from here: https://github.com/draperlaboratory/cbat_tools/pull/244#discussion_r597048015 I'd recommend never having Z3 variables contain `#`. Sadly BIL "nameless" variables by default are of the form e.g. `#123`, so probably hooking `Expr.mk_const_s`, or just `Environment.mk_init_var` would be a good idea.

PR #307, opened 2021-03-22
Address issue #300 by improving Constr.pp_constr, making Constr.t printing more readability

gltrost:
There are several problems currently with printing `Constr.t`-values. Some of these issues are 
1. Escaped special characters like `\n` and `\\`
2. Indentation issues
3. Coloring special-words and values like `let` and `in` when printing 
4. Empty parens `() => ...` which don't need to be shown. 
5. Unreadable syntax like `ITE` instead of `if ... then ... else`

This branch attempts to address or fix all of these issues, mainly by adding changes to `Constr.pp_constr`.

With these changes, reading `Constr.t` statements with flags like `show="precond-internal"` will be much easier.

### DIFF
--- a/wp/lib/bap_wp/src/constraint.ml
+++ b/wp/lib/bap_wp/src/constraint.ml
@@ -51,9 +51,10 @@ let goal_to_string ?colorful:(colorful = false) (g : goal) : string =
   let lhs = g.goal_name |> rem_backslash in 
   let rhs = (Expr.to_string (Expr.simplify g.goal_val None)) |> rem_backslash in
   if colorful then
+    (* lhs is red, rhs is cyan *)
     Format.sprintf "%s%s%s: %s%s%s%!"
       ("\x1b[31m") lhs ("\x1b[0m") ("\x1b[36m") rhs ("\x1b[0m")
-  else Format.sprintf "%s: %s!" lhs rhs
+  else Format.sprintf "%s: %s" lhs rhs
 
   
 let expr_to_hex (exp : z3_expr) : string =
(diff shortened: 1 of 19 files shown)


wp/lib/bap_wp/src/constraint.ml
```diff
@@ -183,25 +190,108 @@ type t =
...
+  let rec rec_pp_constr ch constr =
+    let to_color = to_color colorful ch in
+    match constr with
+    | Goal g ->
+       Format.fprintf ch "%s@;" (goal_to_string ~colorful:colorful g)
+    | ITE (tid, e, c1, c2) ->
+       let color = 35 in (* green *)
+       to_color color "%s: (if " (tid |> Term.tid |> Tid.to_string);
+       Format.fprintf ch "%s" (preen_expr e);
+       to_color color " then%s" "";
```
gltrost, line 261: This is an ugly hack and should be changed. Ideally `to_color` should just take in a color and a format-string, and then the number of inputs for the format-string wouldn't be fixed. The way I've designed it here is that the format-string takes in exactly one input. Thus we end up with these cases where we have `" %s" ""` which are ugly.

jtpaasch (approved):
I mark this as "approved," because my suggestions are just suggestions. Greg, make whichever of my suggested changes you like, and then feel free to merge.

My comments pertain to `constraint.ml`:

* Line 50 et al: maybe change `rem` to `rm`, since `rm` is a common abbreviation for `remove`?
* Line 250: maybe change `form` to `fmt`, since `fmt` is an established abbreviation for formatter arguments? Also, in the `print_expr_list` and `to_color` functions, maybe also use `fmt` instead of `ch` for the formatter arguments? (I can't quite figure out what `ch` is supposed to stand for...)
* Line 250: For what it's worth, I'm a fan of using `t` as the argument name for the constraint. The word `cons` usually means something else in functional programming.
* Line 256: you can use punning here: write `goal_to_string ~colorful g` instead of `goal_to_string ~colorful:colorful g`.
* Line 273: can you use `to_color` here instead of manually adding the escape tags?
* Line 282: can you delete this line?

Other comments:

* I would print the goal names as yellow instead of red. Red is usually used to signal errors in terminal output.
* For the `pp_constr` function, I'd probably make the first argument a named argument `~colorful`. Then in the `constraint.mli` file it is easier to understand what it's for.

Otherwise, looks good! I like how simple you get this stuff boiled down to.

PR #312, opened 2021-04-12
Loop invariant checker

fortunac:
- Adds a flag for checking a loop invariant
  - A user can input an s-expression which represents a list of loop invariants: `(((address 0x12:64u) (invariant <smtlib for invariant>)) (...))`
  - Assumes that a loop only has one entrance and one exit
  - Currently only supported for a single binary
  - If no invariant is specified for a loop, we default to unrolling

- Some changes to existing code:
  - You have to pass in a function for which nodes to skip for `visit_graph`. This was needed to pass in the option to skip nodes in the a loop. `unreachable_from_start` was refactored out, so it is possible to just pass in that.
  - The option to name a constraint created by `mk_smtlib2_single`. This was used to name the invariant "Loop Invariant (tid)" instead of a z3-expression representing the invariant. If no name is specified, uses the default behavior of naming the constraint after the z3 expression.
  - The type for `loop_handler` is changed to hold information about both loop unrolling and loop invariant checking.


wp/lib/bap_wp/src/environment.ml
```diff
@@ -37,6 +37,8 @@ module ExprSet = Set.Make(
     let t_of_sexp _ = raise (Not_implemented "t_of_sexp for z3_expr not implemented")
   end)
 
+let (let*) (x : 'a option) (f : 'a -> 'b) = Option.bind x ~f
```
codyroux, line 40: We might want to move this within a `Utils.Option_Let` module or something. It feels weird to have it at the top level here.

wp/lib/bap_wp/src/environment.ml
```diff
@@ -60,7 +62,7 @@ type t = {
   indirect_handler : indirect_spec;
   jmp_handler : jmp_spec;
   int_handler : int_spec;
-  loop_handler : loop_handler;
+  loop_handler : loop_handle;
```
codyroux, line 65: Why did we change the name?

wp/lib/bap_wp/src/environment.ml
```diff
@@ -177,44 +182,33 @@ let wp_rec_call :
...
   |> Constr.mk_constr
 
-(* Looks up the precondition of the exit node of a loop by:
+(* Looks up the exit node of a loop by:
    - obtaining the post dominator tree
    - for each node in the SCC, find its parent in the dominator tree
    - if the parent node is not in the original SCC, it is an exit node *)
-let loop_exit_pre (env : t) (node : Graphs.Ir.Node.t) (graph : Graphs.Ir.t)
-  : Constr.t option =
+let loop_exit (node : Graphs.Ir.Node.t) (graph : Graphs.Ir.t)
```
codyroux, line 198: It might be nice to summarize these kinds of changes in the PR request, to help my tiny brain wrap around them.

codyroux: I also wonder if the comments couldn't use some upgrades.

wp/lib/bap_wp/src/environment.ml
fortunac: I created a type that holds the different types of handlers we could use for loops. The names were an oversight on my part. I can change the names back.

https://github.com/draperlaboratory/cbat_tools/blob/993075cb5abb04473ac7b82c819cb485b7e694e8/wp/lib/bap_wp/src/environment.ml#L92

https://github.com/draperlaboratory/cbat_tools/blob/993075cb5abb04473ac7b82c819cb485b7e694e8/wp/lib/bap_wp/src/environment.ml#L94

wp/lib/bap_wp/src/environment.ml
```diff
@@ -228,41 +222,54 @@ type unfold_depth = int Unfold_depth.t
      recursively call [Precondition.visit_graph]. Because this function is defined in another
      (later) module, we use open recursion via the [wp_rec_call] function reference. *)
 let rec loop_unfold (num_unroll : int) (depth : unfold_depth) : loop_handler =
```
codyroux, line 224: I have no idea what this change is changing.

wp/lib/bap_wp/src/environment.ml
fortunac: Can do.

wp/lib/bap_wp/src/environment.ml
fortunac: I can update the comments. This function originally found the exit node and returned its precondition. I changed it to return the exit node itself, so the handler can look up the precondition after visiting the subgraph.

wp/lib/bap_wp/src/environment.ml
```diff
@@ -411,9 +420,14 @@ let get_jmp_handler (env : t) : jmp_spec =
...
-let get_loop_handler (env : t) :
-  t -> Constr.t -> start:Graphs.Ir.Node.t -> Graphs.Ir.t -> t =
-  env.loop_handler.handle
+let get_loop_handler (env : t) (tid : Tid.t) : loop_handler =
+  match TidMap.find env.loop_handler.invariant_checker tid with
+  | Some handle ->
+    debug "Checking loop invariant for %s%!" (Tid.to_string tid);
+    handle
+  | None ->
+    debug "Unrolling loop %s%!" (Tid.to_string tid);
```
codyroux, line 429: Should it be "unrolling loop *for* ..." (or maybe "at")?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1388,3 +1396,192 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
   in
   let expr = List.fold l ~init:(Bool.mk_true ctx) ~f:gen_constr in
   Constr.mk_goal "pointer_precond" expr |> Constr.mk_constr
+
+(* Gets the target destination of a jump. We do not handle indirect jumps. *)
+let target (jmp : Jmp.t) : Tid.t option =
```
codyroux, line 1401: I wonder if it would be using this in `visit_jmp` to avoid duplication.

wp/lib/bap_wp/src/environment.ml
fortunac: There's basically no change. I just took the handler out of the record to match the change I made to the `loop_handler` type.

wp/lib/bap_wp/src/precondition.ml
fortunac: I'm confused here. Are you suggesting we use `target` inside of `visit_jmp`?

codyroux: No, just *within* `visit_jmp` to simplify the logic. Not that important.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1388,3 +1396,192 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+            pre, env
+          else
+            visit_elt env pre elt)
+  in
+  let pre, _ = blk |> Blk.elts ~rev:true |> compute in
+  Some pre
+
+(* Gets the body of a loop without the header node containing the loop
+   condition. *)
+let get_body (graph : Graphs.Ir.t) (loop : Graphs.Ir.Node.t group) (blk : Blk.t)
```
fortunac, line 1469: We should probably include the header node in the body (without the backedge into the header).

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1388,3 +1396,192 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+          let unreachable = unreachable_from_start g exit n in
+          in_loop || unreachable
+        in
+        let _, env = visit_graph env post ~start:exit ~skip_node g in
+        Env.get_precondition env tid)
+  in
+  match p with
+  | Some p -> p
+  | None ->
+    let tid = node |> Graphs.Ir.Node.label |> Term.tid in
```
fortunac, line 1544: Add a test where we would reach this point where we can't find the precondition.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1388,3 +1396,192 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+  let visitor =
+    (object inherit [Var.Set.t] Term.visitor
+      method! visit_def def vars =
+        Var.Set.add vars (Def.lhs def)
+    end)
+  in
+  visitor#visit_sub sub Var.Set.empty
+
+(* Creates fresh names for [vars] found in the constraint. *)
+let freshen (constr : Constr.t) (env : Env.t) (vars : Var.Set.t)
```
fortunac, line 1517: Move this to be next to `subst_fun_output` and have `spec_verifier_nondet` call this.

wp/lib/bap_wp/src/constraint.ml
```diff
@@ -208,35 +208,35 @@ let rec del_empty_constr_hyps (const : t) : t =
     let fold_list c_list =
       List.fold_right c_list ~init:[]
         ~f:(fun x y ->
-          match x with
-          | Clause([],xs) ->
-             List.append (List.map xs ~f:del_empty_constr_hyps) y
-          | _ -> (del_empty_constr_hyps x) :: y) in
+            match x with
```
codyroux, line 211: Are we ocp-indent compliant?

fortunac: I have a git hook that runs ocp-indent on all the files I edit. That's the source of all these whitespace changes.

codyroux: Great, I'll just ignore whitespace from now on in the commits.

wp/lib/bap_wp/src/compare.ml
```diff
@@ -169,15 +169,17 @@ let compare_subs
...
+    let post = Constr.trivial ctx in
     post, env1, env2
   in
   let hyps = postcond in
   postcond, hyps
 
 let compare_subs_empty_post : comparator * comparator =
   let postcond ~original:(_, env1) ~modified:(_, env2) ~rename_set:_ =
-    let post = Env.trivial_constr env1 in
+    let ctx = Env.get_context env1 in
```
codyroux, line 181: ?

wp/lib/bap_wp/src/environment.ml
```diff
@@ -168,102 +172,20 @@ let init_sub_handler (subs : Sub.t Seq.t) (target : Theory.target)
...
-
-(* This is the default handler for loops, which unfolds a loop by:
-   - Looking at the target node for a backjump
-   - If the node has been visited more than [num_unroll] times, use the [loop_exit_pre] precondition
-   - Otherwise, decrement the [depth] map which tracks the unfoldings for that node, and
-     recursively call [Precondition.visit_graph]. Because this function is defined in another
-     (later) module, we use open recursion via the [wp_rec_call] function reference. *)
-let rec loop_unfold (num_unroll : int) (depth : unfold_depth) : loop_handler =
+let init_loop_handler
+    ~(default : t -> Constr.t -> start:Graphs.Ir.Node.t -> Graphs.Ir.t -> t)
```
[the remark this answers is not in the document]
codyroux: (not sure if we should just unbox that type)

wp/lib/bap_wp/src/environment.ml
```diff
@@ -168,102 +172,20 @@ let init_sub_handler (subs : Sub.t Seq.t) (target : Theory.target)
...
+    (handlers : (Tid.t -> (t -> Constr.t -> start:Graphs.Ir.Node.t ->
```
codyroux, line 177: Same comment.

wp/lib/bap_wp/src/environment.ml
```diff
@@ -281,21 +203,23 @@ let init_loop_unfold (num_unroll : int) : loop_handler = loop_unfold num_unroll
...
-    ~exp_conds:(exp_conds : exp_cond list)
-    ~num_loop_unroll:(num_loop_unroll : int)
-    ~target:(target : Theory.target)
-    ~freshen_vars:(freshen_vars : bool)
-    ~use_fun_input_regs:(fun_input_regs : bool)
-    ~stack_range:(stack_range : mem_range)
-    ~data_section_range:(data_section_range : mem_range)
-    ~func_name_map:(func_name_map : string StringMap.t)
-    ~smtlib_compat:(smtlib_compat : bool)
+    ~(subs : Sub.t Seq.t)
```
codyroux, line 206: This syntax is indeed nicer!

codyroux: Though it makes it hard to see what actually changed...

wp/lib/bap_wp/src/compare.ml
fortunac: I moved the trivial constraint from `environment.ml` to `constraint.ml` because this was used in the loop unroller and since I moved the loop unroller to `precondition.ml`, keeping that function in `environment.ml` would cause cyclical dependencies. I also think trivial fits `constraint.ml` better than `environment.ml`.

wp/lib/bap_wp/src/environment.ml
```diff
@@ -523,3 +448,9 @@ let get_init_var (env : t) (var : Var.t) : Constr.z3_expr option =
 
 let map_sub_name (env : t) (name_mod : string) : string =
   get_mapped_name name_mod env.func_name_map
+
+let get_unroll_depth (env : t) : unroll_depth =
+  env.unroll_depth
+
+let set_unroll_depth (env : t) (depth : unroll_depth) : t =
```
codyroux, line 455: do we want to wrap this so that `set_unroll_depth` just takes a `(key, value)` pair (or whatever it actually needs)?

wp/lib/bap_wp/src/environment.ml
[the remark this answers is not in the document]
fortunac: Are you referring to how the `loop_handler` type is currently a record with the `handle` field?

wp/lib/bap_wp/src/environment.ml
codyroux: On one hand, I guess a client could want to just run amok on this entry, on the other hand, maybe we want to constrain the client to just do what you're expected to do, which is decrement an unfold.

Not sure what is better.

docs/reference.html
```diff
@@ -116,6 +116,8 @@ <h2>Single program analysis</h2>
...
 
 <li><code>--postcond=&lt;smt-lib-string&gt;</code> &mdash; Allows you to specify an assertion that WP will assume is true at the end of the function it is analyzing, using the <a href="#smt-lib">smt-lib2 format</a>. If no postcondition is specified, a trivial postcondition of <code>true</code> will be used. This flag can also be used in comparative analysis; see <a href="#comparative">below</a>.</li>
+
+<li><code>--loop-invariant=&lt;s-expression&gt;</code> &mdash; Usage: <code>(((address &lt;addr&gt;) (invariant &lt;smtlib&gt;)) (...))</code>. Assumes the subroutine contains unnested loops with one entrance and one exit each. Checks the loop invariant written in smt-lib2 format for the loop with its header at the given address. The address should be written in BAP's bitvector string format. Only supported for a single binary analysis.</li>
```
codyroux, line 120: can the loop be in either `do ... while(foo)` or `while(foo){ ... }` or can we handle both?

codyroux: (I'd say "entry point" rather than "entrance")

wp/lib/bap_wp/src/environment.ml
fortunac: Are you referring to how we have

```
  let find_depth depth node = Env.Unroll_depth.find depth (Node.label node)
                              |> Option.value ~default:num_unroll
  in
  let decr_depth depth node =
    Env.Unroll_depth.update depth (Node.label node)
      ~f:(function None -> num_unroll - 1 | Some n -> n - 1)
  in
```
separately, and can make the api based on these rather than giving the client full reign over the map itself?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -398,12 +399,21 @@ let subst_fun_outputs ?tid_name:(tid_name = "") ~inputs:(inputs : Var.t list)
       Env.add_call_pred env sub_to) in
   Constr.substitute post subs_from subs_to, env
 
+(* Creates fresh names for [vars] found in the constraint. Optionally takes in
+   a function that maps the original variable name to the fresh name. Defaults
+   to [fresh_var]. *)
+let freshen ?(name = Format.sprintf "fresh_%s") (constr : Constr.t)
```
codyroux, line 405: I wonder if this should be in `environment.ml`; it seems useful.

docs/reference.html
fortunac: I'll need to test more...

wp/lib/bap_wp/src/precondition.ml
fortunac: Can move it there!

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -398,12 +399,21 @@ let subst_fun_outputs ?tid_name:(tid_name = "") ~inputs:(inputs : Var.t list)
...
+    (env : Env.t) (vars : Var.Set.t) : Constr.t * Env.t =
+  Var.Set.fold vars ~init:(constr, env) ~f:(fun (constr, env) v ->
+      let z3_v, env = Env.get_var env v in
+      let name = name (Expr.to_string z3_v) in
+      let fresh = Env.new_z3_expr ~name env (Var.typ v) in
+      Constr.substitute_one constr z3_v fresh, Env.add_call_pred env fresh)
```
codyroux, line 411: This gonna sound silly, but any reason we don't do them all at once?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -607,11 +617,8 @@ let spec_verifier_nondet (sub : Sub.t) (_ : Theory.target) : Env.fun_spec option
                | Some o -> o
                | None -> failwith "Verifier headerfile must be specified with --api-path" in
              let vars = output |> Bap.Std.Arg.rhs |> Exp.free_vars in
-             let v = Var.Set.choose_exn vars in
-             let z3_v, env = Env.get_var env v in
-             let name = Format.sprintf "%s_ret_%s" (Sub.name sub) (Expr.to_string z3_v) in
-             let fresh = Env.new_z3_expr env ~name:name (Var.typ v) in
-             Constr.substitute_one post z3_v fresh, Env.add_call_pred env fresh)
+             let name = Format.sprintf "%s_ret_%s" (Sub.name sub) in
+             freshen ~name post env vars)
```
codyroux, line 621: 👍

wp/lib/bap_wp/src/environment.ml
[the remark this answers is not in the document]
codyroux: Yes, I'm not sure what it's buying us.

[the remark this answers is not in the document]
codyroux: It's a common Haskell pattern because you can ensure that the type doesn't get boxed using `newtype`.

wp/lib/bap_wp/src/precondition.ml
fortunac: `Env.add_call_pred` was added a long time after `Constr.substitute_one`. It was added so that variables that are introduced during the WP would also be printed out in the model (i.e. in the function specs and here in the loop invariant checker). I'm not sure we want to combine the two because there are many occasions where we substitute an expression without introducing a new variable.

codyroux: I just mean use `Constr.substitute` on all the vars at once, instead of folding and doing a bunch of `Constr.substitute_one`.

Maybe we optimize this at the end when building the Z3 constraint?

fortunac: Ooh! I have definitely use `Constr.substitute` instead.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -916,21 +886,23 @@ let visit_block (env : Env.t) (post : Constr.t) (blk : Blk.t) : Constr.t * Env.t
...
+    ~(start : Graphs.Ir.Node.t) ~(skip_node : Graphs.Ir.Node.t -> bool)
+    (g : Graphs.Ir.t) : Constr.t * Env.t =
   let module G = Graphs.Ir in
-  (* Prune non-reachable sub-graphs from the DFS *)
-  let module Reachable_from_start =
-    (val
-      Graphlib.filtered (module G)
-        ~skip_node:(fun n -> not (Graphlib.is_reachable (module G) g start n)) ())
-  in
+  let module Filtered_graph = (val Graphlib.filtered (module G) ~skip_node ()) in
```
codyroux, line 899: 👍

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -941,17 +913,12 @@ let visit_graph (env : Env.t) (post : Constr.t)
         let handler = Env.get_loop_handler env in
         post, handler env post ~start:dst g
       end
-    | _ -> p
+    | _ -> post, env
```
codyroux, line 916: again, nice catch. Maybe worth leaving a comment here about why this `post` is the good `post`.

As a sign `post` for other confused readers.

fortunac: 👍

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -916,21 +886,23 @@ let visit_block (env : Env.t) (post : Constr.t) (blk : Blk.t) : Constr.t * Env.t
   let pre, env = blk |> Blk.elts ~rev:true |> compute_pre in
   (pre, Env.add_precond env (Term.tid blk) pre)
 
+(* Returns [true] if the node is not reachable from the start of the graph.
+   This is used to prune non-reachable subgraphs from the DFS. *)
+let unreachable_from_start (graph : Graphs.Ir.t) (start : Graphs.Ir.Node.t)
+    (node : Graphs.Ir.Node.t) : bool =
+  not (Graphlib.is_reachable (module Graphs.Ir) graph start node)
+
 let visit_graph (env : Env.t) (post : Constr.t)
```
codyroux, line 895: Do we need comments here? It's worth noting that if you skip nodes for which another node needs the precondition, this function may fail.

fortunac: Can do.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1425,3 +1392,288 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+      if Group.mem group parent then
+        None
+      else
+        Some parent)
+
+(* This is the default handler for loops, which unrolls a loop by:
+   - Looking at the target node for a backjump
+   - If the node has been visited more than [num_unroll] times, use the [loop_exit_pre] precondition
+   - Otherwise, decrement the [depth] map which tracks the unrollings for that node, and
+     recursively call [Precondition.visit_graph]. Because this function is defined in another
```
codyroux, line 1419: Looks like this last sentence is now moot? We should remove it if so.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1425,3 +1392,288 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+
+(* Assumes the loop header branches into two blocks. *)
+let get_branch (blk : Blk.t) : (Jmp.t * Jmp.t) option =
+  match Seq.to_list (Term.enum jmp_t blk) with
+  | [jmp1; jmp2] -> Some (jmp1, jmp2)
+  | _ -> None
+
+(* Gets the BAP expression that represents the condition for going into the
+   loop. *)
+let get_branch_condition (loop : Graphs.Ir.Node.t group) (jmp1 : Jmp.t)
```
codyroux, line 1486: Still a bit confused if we're handling this correctly: we need to apply whatever's in this block to the condition as well, right?

codyroux: If that's the case, then I feel like a clever version of `visit_blk` should be used instead of this function.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1425,3 +1392,288 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+    else if in_loop loop target2 then
+      (* jmp1 exits, jmp2 goes to loop. while(E) is the negation of jmp1's
+         condition. *)
+      Some Bil.(unop not (Jmp.cond jmp1))
+    else
+      None
+  | _, _ -> None
+
+(* Gets a constraint representing the condition for a loop. *)
+let get_cond (loop : Graphs.Ir.Node.t group) (blk : Blk.t) (env : Env.t)
```
codyroux, line 1502: Again, this feels like code duplication from `visit_blk`. I wonder if we can generalize it to handle both cases.

wp/lib/bap_wp/src/precondition.ml
codyroux: Indeed.

wp/lib/bap_wp/src/precondition.ml
fortunac: The `compute` in `get_cond` is basically the same as `visit_blk` except that it skips jumps, so I can probably do something similar to what I did with `visit_graph` and pass in a function that determines which instructions to skip.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1425,3 +1392,288 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+      Graphs.Ir.Edge.remove edge g
+    else
+      g
+  in
+  Graphlib.depth_first_search (module Graphs.Ir)
+    ~enter_node ~enter_edge ~init:graph graph
+
+(* Splits the loop into the condition and the body of the loop. Assumes there
+   are no breaks in the loop. *)
+let split_loop (loop : Graphs.Ir.Node.t group) (graph : Graphs.Ir.t) (env : Env.t)
```
codyroux, line 1557: Is it possible to have a dummy precondition for the loop exit to  avoid having to do this split?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1425,3 +1392,288 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+  let sub = Sub.of_cfg graph in
+  Seq.find_map (Term.enum blk_t sub) ~f:(fun blk ->
+      match get_cond loop blk env with
+      | Some cond ->
+        let body = Sub.of_cfg (get_body graph loop blk) in
+        Some (cond, body)
+      | None -> None)
+
+(* Gets a set of vars that get updated in the body of a loop. *)
+let loop_vars (sub : Sub.t) : Var.Set.t =
```
codyroux, line 1568: 👍

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1425,3 +1392,288 @@ let construct_pointer_constraint (l_orig : Constr.z3_expr list) (env1 : Env.t)
...
+        Env.get_precondition env tid)
+  in
+  match p with
+  | Some p -> p
+  | None ->
+    let tid = node |> Graphs.Ir.Node.label |> Term.tid in
+    warning "Trivial precondition is being used for node %s\n%!" (Tid.to_string tid);
+    post
+
+let loop_invariant_checker (loop_invariants : string Tid.Map.t) (tid : Tid.t)
```
codyroux, line 1600: Nice and clean!

codyroux: Maybe we should just be honest and link to the wikipedia page.

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -285,7 +286,10 @@ val mk_env
   -> ?jmp_spec:Env.jmp_spec
   -> ?int_spec:Env.int_spec
   -> ?exp_conds:Env.exp_cond list
-  -> ?num_loop_unroll:int
+  -> ?loop_handlers:((Bap.Std.Tid.t -> (Env.t -> Constr.t -> start:Bap.Std.Graphs.Ir.Node.t ->
```
codyroux, line 289: Same comment about compressing this with a `loop_handler` type.

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -364,4 +368,21 @@ val set_sp_range : Env.t -> Constr.t
     all specified registers cannot point to the uninitalized stack region.
     They must be either below the bottom of the stack or above the initial
     stack pointer. *)
+
 val construct_pointer_constraint : Constr.z3_expr list -> Env.t -> (Constr.z3_expr list) option -> Env.t option -> Constr.t
+
+(** [loop_unroll num_unroll] returns a function that will unroll loops
```
codyroux, line 374: Maybe be more explicit: "will compute the precondition of nodes within loops by assuming that the loop was traversed at most `num_unroll` times".

Or something. Not too big an issue.

wp/lib/bap_wp/src/precondition.mli
```diff
@@ -364,4 +368,21 @@ val set_sp_range : Env.t -> Constr.t
...
+val loop_unroll
+  : int ->
+  (Env.t -> Constr.t -> start:Bap.Std.Graphs.Ir.Node.t
+   -> Bap.Std.Graphs.Ir.t -> Env.t)
+
+(** [loop_invariant_checker invariants tid] returns a function that will check
+    the loop invariant for the loop that starts at [tid] given that the
+    invariant exists in [invariants]. *)
+val loop_invariant_checker
+  : string Bap.Std.Tid.Map.t
```
codyroux, line 385: Should we just create a type alias for this? Or name the argument?

wp/lib/bap_wp/src/run_parameters.ml
```diff
@@ -56,6 +58,13 @@ type t = {
   ext_solver_path : string option
 }
 
+type loop_invariant = {
```
codyroux, line 61: 👍

wp/lib/bap_wp/src/precondition.mli
fortunac: Something like `type loop_invariants = string Tid.Map.t`?

wp/lib/bap_wp/src/run_parameters.ml
```diff
@@ -56,6 +58,13 @@ type t = {
...
+  address : string;
```
codyroux, line 62: is this a hex address, or a tid?

fortunac: Hex address in the form `(address 0x12:64u)`. I couldn't figure out how to use the tid labels from the KB.

wp/lib/bap_wp/src/run_parameters.ml
```diff
@@ -153,6 +162,27 @@ let validate (f : t) (files : string list) : (unit, error) result =
...
 
+(* Parses the loop invariant and address from the user into the format accepted
+   by the environment. *)
+let parse_loop_invariant (invariants : string) (sub : Sub.t)
+  : string Tid.Map.t =
+  if String.is_empty invariants then
+    Tid.Map.empty
+  else
+    Term.enum blk_t sub
+    |> Seq.fold ~init:Tid.Map.empty ~f:(fun map blk ->
```
codyroux, line 173: I wonder if there's an easier way to do this...

Probably not (for now).

wp/lib/bap_wp/src/run_parameters.ml
```diff
@@ -153,6 +162,27 @@ let validate (f : t) (files : string list) : (unit, error) result =
...
+        match Term.get_attr blk address with
+        | None -> map
+        | Some address ->
+          let invs = invariant_list_of_sexp (Sexp.of_string invariants) in
+          List.fold invs ~init:map ~f:(fun m inv ->
+              let addr = Addr.of_string inv.address in
+              if Addr.equal address addr then
```
codyroux, line 180: Should we set a boolean here and print a warning if the address is *not* found?

fortunac: Sounds like a good idea.

wp/lib/bap_wp/src/run_parameters.mli
```diff
@@ -97,5 +99,11 @@ val validate_mem_flags : bool -> bool -> string list -> (unit, error) result
     when there are two files to compare. Returns an error otherwise. *)
 val validate_check_invalid_derefs : bool -> string list -> (unit, error) result
 
+(** [parse_loop_environment invariant sub] parses the [invariant] which is an
+     S-expression representing the address of a loop header and its
+     corresponding loop invariant, and returns a format accepted by the
+     environment. *)
```
codyroux, line 105: Have an example here?

wp/lib/bap_wp/src/utils.ml
```diff
@@ -17,12 +17,15 @@ open Bap.Std
...
 include Self()
 
-module Pre = Precondition
+module Option_let = struct
 
-let update_default_num_unroll (num_unroll : int option) : unit =
-  match num_unroll with
-  | Some n -> Pre.num_unroll := n
-  | None -> ()
+  let (let*) (x : 'a option) (f : 'a -> 'b option) : 'b option =
```
codyroux, line 22: +1

wp/lib/bap_wp/tests/unit/test_utils.ml
```diff
@@ -20,32 +21,33 @@ module Utils = Bap_wp.Utils
...
-    | Some n ->
-      let fail_msg =
-        Format.sprintf "num_unroll was not updated from %d to %d" original n in
-      assert_bool fail_msg (original <> updated)
-    | None ->
-      let fail_msg =
-        Format.sprintf "Num unroll was updated from %d but should have been \
-                        unchanged" original in
-      assert_equal ~ctxt:ctxt ~cmp:Int.equal ~msg:fail_msg updated original
+let test_get_mod_func_name
```
codyroux, line 24: 💯

wp/plugin/wp.ml
```diff
@@ -288,8 +297,8 @@ let grammar = Cmd.(
     $ func_name_map
     $ user_func_spec
     $ fun_specs
-    $ files
-    $ ext_solver_path)
+    $ ext_solver_path
```
codyroux, line 300: ?

wp/resources/sample_binaries/loop_invariant/in_registers/src/main.c
```diff
@@ -0,0 +1,16 @@
+#include <assert.h>
+
+int loop(void) {
+  register int x asm ("rdi") = 0;
```
codyroux, line 4: Dang! Cool.

wp/resources/sample_binaries/loop_invariant/on_stack/loop_invariant.smt
```diff
@@ -0,0 +1,14 @@
+(((address 0x12:64u)
+ (invariant
+   "(assert
+      (let ((x!1 (concat (select mem (bvadd #xffffffffffffffff RBP))
```
codyroux, line 4: Oof. I wonder if we can use this `let` trick.

codyroux:
Absolutely stellar code!

My comments are a bit of an afterthought: it feels like we're duplicating a bit of the `visit_blk` functionality, it would be nice to unify that code somehow.

We can leave that as a TODO if a solution is not forthcoming.

wp/lib/bap_wp/src/compare.ml
codyroux: Ok, makes sense!

wp/lib/bap_wp/src/environment.ml
codyroux: Yes, though I'm not sure this is the right thing to do.

What's your feeling?

wp/lib/bap_wp/src/precondition.ml
codyroux: Might be the right thing. Let's discuss maybe.

wp/lib/bap_wp/src/precondition.ml
codyroux: We should discuss, I'm not sure.

wp/lib/bap_wp/src/precondition.mli
codyroux: Yep.

wp/lib/bap_wp/src/run_parameters.ml
codyroux: Can we infer the `:64u`? I suspect users may not want to type that.

ccasin:
I am playing around with this and it is cool!

One thing that might be nice to start doing: documentation in our examples/test cases.  A README.md file in each subdirectory under `wp/resources/sample_binaries/loop_invariant` that just says in English what the test is supposed to check and what the property means would be helpful.

fortunac:
> 
> 
> I am playing around with this and it is cool!
> 
> One thing that might be nice to start doing: documentation in our examples/test cases. A README.md file in each subdirectory under `wp/resources/sample_binaries/loop_invariant` that just says in English what the test is supposed to check and what the property means would be helpful.

Can definitely do this!

PR #313, opened 2021-04-14
Codyroux/arch to target

codyroux:
Implement a rather large refactoring, to use Theory.target instead of Arch.t everywhere.

This is more in line with BAP 2.0 philosophy, but also is required to handle the new Thumb target.


wp/lib/bap_wp/src/environment.ml
```diff
@@ -62,7 +63,7 @@ type t = {
   int_handler : int_spec;
   loop_handler : loop_handler;
   exp_conds : exp_cond list;
-  arch : Arch.t;
+  arch : Theory.target;
```
fortunac, line 66: Is it possible to change the name of this field to `target`?

wp/lib/bap_wp/src/environment.ml
```diff
@@ -418,29 +419,37 @@ let get_loop_handler (env : t) :
 let get_call_preds (env : t) : ExprSet.t =
   env.call_preds
 
-let get_arch (env : t) : Arch.t =
+let get_arch (env : t) : Theory.target =
   env.arch
 
+let var_comp : (_, _) Set.comparator = (module Var.Set.Elt)
```
fortunac, line 425: What is `var_comp` supposed to do? Maybe a comment would help.

wp/lib/bap_wp/src/environment.ml
```diff
@@ -455,7 +464,7 @@ let get_smtlib_compat (env : t) : bool = env.smtlib_compat
 let init_stack_ptr (env : t) : Constr.z3_expr -> Constr.z3_expr =
   let ctx = get_context env in
   let arch = get_arch env in
-  let sort = arch |> Arch.addr_size |> Size.in_bits |> BV.mk_sort ctx in
+  let sort = arch |> Theory.Target.data_addr_size |> BV.mk_sort ctx in
```
fortunac, line 467: Would it be a good idea to add a comment here so we know this is in bits?

wp/lib/bap_wp/src/output.ml
```diff
@@ -200,11 +200,15 @@ let output_gdb (solver : Solver.solver) (status : Solver.status)
     let model = Constr.get_model_exn solver in
     let mem_model = get_mem model env in
     let varmap = Env.get_var_map env in
-    let module Target = (val target_of_arch (Env.get_arch env)) in
-    let regmap = VarMap.filter_keys ~f:(Target.CPU.is_reg) varmap in
+    let target = (Env.get_arch env) in
+    let to_core v = Theory.Var.create (Var.sort v) (Var.ident v) in
+    (* FIXME: does this return [mem] by accident? *)
```
fortunac, line 205: Can you clarify this section? Is `to_core` the one returning `mem`, and is `is_reg` properly filtering it out?

wp/lib/bap_wp/src/output.ml
```diff
@@ -221,11 +225,14 @@ let output_bildb (solver : Solver.solver) (status : Solver.status) (env : Env.t)
   match status with
   | Solver.SATISFIABLE ->
     info "Outputting BilDB init script to %s\n%!" filename;
-    let module Target = (val target_of_arch (Env.get_arch env)) in
     let model = Constr.get_model_exn solver in
     let mem_model = get_mem model env in
     let var_map = Env.get_var_map env in
-    let reg_map = VarMap.filter_keys ~f:(Target.CPU.is_reg) var_map in
+    let target = (Env.get_arch env) in
```
fortunac, line 231: Would it be a good idea to refactor this out so we don't have duplicated code?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -303,14 +304,16 @@ let set_fun_called (post : Constr.t) (env : Env.t) (tid : Tid.t) : Constr.t =
...
   let arch = Env.get_arch env in
   if Env.is_x86 arch then
     begin
-      let module Target = (val target_of_arch arch) in
-      let sp, env = Env.get_var env Target.CPU.sp in
-      let width = Target.CPU.sp |> Var.typ |> typ_size in
-      let addr_size = arch |> Arch.addr_size |> Size.in_bytes in
+      let sp, env = Env.get_sp env |> Env.get_var env in
+      let width = arch |> Theory.Target.bits in
+      (* let addr_size = arch |> Theory.Target.data_addr_size in *)
```
fortunac, line 314: Can we delete this section of commented out code?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -396,62 +399,84 @@ let subst_fun_outputs ?tid_name:(tid_name = "") ~inputs:(inputs : Var.t list)
...
-  | #Arch.arm ->
-    let open ARM.CPU in
-    [r4; r5; r6; r7; r8; r9; r10; r11]
-  | a ->
-    warning "callee_saved_regs: Callee-saved registers have not \
-             been implemented for %s." (Arch.to_string a);
-    []
+(* FIXME: use built-in BAP roles? *)
+let input_regs (arch : Theory.target) : Var.t list =
+  if (Theory.Target.matches arch "amd64") then
```
fortunac, line 404: Having magic strings all over these functions makes me a bit nervous.  Can we define the targets outside of the function and use those variables as identifiers instead? We are matching on the strings explicitly in many places in `Environment.ml` and `Precondition.ml`.

```
let amd64 = "amd64"

...

if (Theory.Target.matches arch amd64) then ...
```

fortunac (commented):
This is a very impressive refactor! Most of my comments are nitpicky, but otherwise everything looks good!

wp/lib/bap_wp/src/environment.ml
codyroux: Maybe? This feels like the responsibility of the BAP API, not us.

Ivan claims that bits is better all arounds, since apparently there are systems for which a byte is not 4 bits (crazy).

https://stackoverflow.com/a/18743555/2059388

wp/lib/bap_wp/src/output.ml
codyroux: `to_core` is just a type cast from `Bap.Std.Var.t` to `Theory.Var.t`.

I'll clarify what this code is doing in comments.

wp/lib/bap_wp/src/output.ml
codyroux: Agreed.

wp/lib/bap_wp/src/precondition.ml
codyroux: Yes!

wp/lib/bap_wp/src/precondition.ml
codyroux: One might argue that these strings are not really magic at all, since they represent exactly what they mean.

I agree that that bap likes to use `amd64` rather than `x86_64` is somewhat non-obvious, but I'm not sure how to resolve this.

We can hoist this out into a single let-binding, if you like.

wp/lib/bap_wp/src/environment.ml
fortunac: Sounds good. We can keep this if you feel it's better.

wp/lib/bap_wp/src/precondition.ml
codyroux: On reflection, this would probably be cleaner with `is_amd64`, `is_arm` etc. functions

fortunac: That sounds good! We already have an `is_x86` so maybe it would be a good idea to have for `arm` etc.

I just want to avoid the situation where someone writes something like `Theory.Target.matches arch "amd644"` and we don't catch it at compile time.

codyroux: Reasonable.

wp/lib/bap_wp/src/environment.ml
codyroux: Sure. FYI `var_comp` is the comparator for `Var` sets, which I need to cast sets of `Theory.Var.Top.t`s into `Bap.Std.Var.t`s. A bit of tedious book-keeping, really.

wp/lib/bap_wp/src/environment.ml
codyroux: Tedious (it changes the API), but possible.

Unless you don't want to change the name of the `mk_env` argument, but then what's the point?

codyroux: I'll change it.

wp/lib/bap_wp/src/environment.ml
codyroux: I don't think this warrants a comment, since it's making use of the API as intended.

PR #315, opened 2021-04-20
External SMT Model Keyword "Model" Bug fix

philzook58:
Relaxed condition looking for keyword "model" in external smt model and enhanced error messages for easier later debugging.

### DIFF
--- a/wp/lib/bap_wp/src/z3_utils.ml
+++ b/wp/lib/bap_wp/src/z3_utils.ml
@@ -166,8 +166,9 @@ let asserts_of_model (model_string : string) (sym_names : string list) : Sexp.t 
             model_val
           | Sexp.List _  ->  (* Function model *)
             Sexp.List [Sexp.Atom "lambda" ; args; model_val] (* Sexp.Atom varname *)
-          | Sexp.Atom a -> failwith (sprintf "Unexpected atom %s in external model %s"
-                                     a model_string)
+          | Sexp.Atom a -> failwith (sprintf "model_string: %s\n
+                  function asserts_of_model: Unexpected atom %s in external model\n"
+                                     model_string a)
         in
         Sexp.List [Sexp.Atom "assert" ;
                    Sexp.List [Sexp.Atom "=" ; Sexp.Atom varname ; model_val]]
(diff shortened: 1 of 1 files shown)


wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -175,13 +176,15 @@ let asserts_of_model (model_string : string) (sym_names : string list) : Sexp.t
...
+    | _ -> failwith (sprintf "Error: Unexpected form in external smt model: %s\n"
+                     model_string)
   in
   let model_sexp = Sexp.of_string model_string in
   match model_sexp with
-  | Sexp.List (Sexp.Atom "model" :: t) ->
+  | Sexp.List (Sexp.Atom "model" :: t) | Sexp.List t ->
     List.map ~f:process_decl t
-  | _ -> failwith "Error: Unexpected form in external smt model"
+  | _ -> failwith (sprintf "Error: Unexpected form in external smt model: %s\n"
```
jtpaasch, line 186: Can we make this message different from the one on line 179?

codyroux: Also, I've been using the informal convention `failwith "function_name: error message"` I dunno if we want this though.

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -166,7 +166,8 @@ let asserts_of_model (model_string : string) (sym_names : string list) : Sexp.t
             model_val
           | Sexp.List _  ->  (* Function model *)
             Sexp.List [Sexp.Atom "lambda" ; args; model_val] (* Sexp.Atom varname *)
-          | Sexp.Atom _ -> failwith "Unexpected atom"
+          | Sexp.Atom a -> failwith (sprintf "Unexpected atom %s in external model %s"
```
codyroux, line 169: Do we want to print the whole model here? I wonder if it just drowns the user.

codyroux:
Sorry for the nits, Wadler's law I guess.

Looks fine otherwise.

wp/lib/bap_wp/src/z3_utils.ml
philzook58: We do. In the case this error message gets printed I need someone to go in and add print statements so I can see the model to know what went wrong.

philzook58: I guess I could put the error message after the model so it isn't as buried?

codyroux: All right.

jtpaasch: If we're forcing a fail with a message, then it's clearly not a problem for the average user. This event itself will "drown" the average user. For these kind of hard crashes that users can't really handle themselves, I think it's best to print the problem fully, so that the user who encounters it can just cut and paste the data into a github issue.

codyroux:
Seems fine, merging this because it's somewhat urgent.

PR #321, opened 2021-05-21
Fortunac/general loop invariant checker

fortunac:
- Simplifies the loop invariant checker to handle more general cases
  - Rather than finding the loop guard and creating constraints from those, we find the weakest precondition for the loop itself.
  - `invariant /\ (invariant => ite(!loop_guard, precondition of the exit, wp(loop_body, invariant))`
- Adds more test cases:
  - A loop with a nondeterministic number of iterations
  - A do while loop
  - A loop with a break
  - A non-terminating loop


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -901,8 +901,12 @@ let visit_graph (env : Env.t) (post : Constr.t)
         let dst = G.Edge.dst e in
         debug "Entering back edge from\n%sto\n%s\n%!"
           (G.Node.to_string src) (G.Node.to_string dst);
-        let handler = Env.get_loop_handler env in
-        post, handler env post ~start:dst g
+        let tid = dst |> G.Node.label |> Term.tid in
```
codyroux, line 904: I'm wondering, is the general approach of how we handle this documented anywhere?

I guess this is more of a new issue than anything else.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1397,47 +1401,36 @@ let in_loop (loop : Graphs.Ir.Node.t group) (tid : Tid.t) : bool =
    - for each node in the SCC, find its parent in the dominator tree
    - if the parent node is not in the original SCC, it is an exit node.
 
-   The exit node of a loop is the first node outside of the SCC in the control
-   flow graph. This should be the node the loop header points to that isn't a
-   part of the loop.
+   The exit node of a loop is a node outside of the SCC in the control flow
+   graph. This node will always be visited when exiting a loop.
```
codyroux, line 1405: Not sure what that last sentence means. Is it "will have already been visited"?

wp/lib/bap_wp/src/precondition.ml
fortunac: I could either put this in the comments for the loop invariant checker or in the design decisions markdown file. Leaning towards the design decisions.

wp/lib/bap_wp/src/precondition.ml
fortunac: The execution of the program will always go through the exit node if we are on the path that contains the loop.

codyroux: Is this true though? What if a loop has 2 exits?

fortunac: Hmm.. I guess not always true. I only had a few samples with multiple exits, and this always returned the node that all the exits pointed to. I can't say for other loop shapes.

codyroux: So we don't actually know what this code does?

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1397,47 +1401,36 @@ let in_loop (loop : Graphs.Ir.Node.t group) (tid : Tid.t) : bool =
...
+  Seq.find_map (Group.enum loop) ~f:(fun n ->
       let* parent = Tree.parent dom_tree n in
-      if Group.mem group parent then
+      if Group.mem loop parent then
         None
       else
         Some parent)
 
 (* Gets the precondition of the exit node of a loop. *)
 let exit_pre (env : Env.t) (post : Constr.t) (node : Graphs.Ir.Node.t)
```
codyroux, line 1420: 👍

codyroux: Though I guess we could do this *as we hit the nodes* during the loop analysis, but I'm not sure that would improve the code.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1468,127 +1466,27 @@ let loop_unroll (num_unroll : int) : Env.loop_handler =
...
-  Graphlib.depth_first_search (module Graphs.Ir)
-    ~enter_node ~enter_edge ~init:graph graph
-
-(* Gets a set of vars that get updated in the body of a loop. *)
-let loop_vars (sub : Sub.t) : Var.Set.t =
-  let visitor =
-    (object inherit [Var.Set.t] Term.visitor
-      method! visit_def def vars =
-        Var.Set.add vars (Def.lhs def)
-    end)
```
codyroux, line 1580: Oof that feels nice

codyroux:
This is neato burrito! Very nice stuff.

wp/lib/bap_wp/src/precondition.ml
fortunac: Are you suggesting a hook in `visit_graph` where we calculate the parent as we enter the node, and then if it is a parent in the post-dominator tree, we recursively call `visit_graph` to update the env?

codyroux: Yes, that's what I'm suggesting, as maybe a `jmp_spec` or whatever we call it.

Again: not sure that would make the code cleaner in any way, so I wouldn't sweat it.

PR #336, opened 2021-08-11
Init Rodata

codyroux:
This is ~~not~~ ready for merge yet, ~~I'm just creating a PR to telegraph the review, allow discussion and prepare for the merge.~~

Seems about ready! I think I may have forgotten to add the system test, but it's written...


wp/plugin/lib/wp_utils.ml
```diff
@@ -53,22 +52,38 @@ let clear_mapper : Term.mapper = object
...
+  let project_digest = Cache.Digests.project mk_digest in
+  match Project.Cache.load project_digest with
+  | Some project ->
     info "Program %s (%a) found in cache.%!"
-      filepath Data.Cache.Digest.pp program_digest;
-    prog
+      filepath Data.Cache.Digest.pp project_digest;
+    project
   | None ->
     (* The program_t is not in the cache. Disassemble the binary. *)
```
fortunac, line 77: Can we update this comment?

wp/plugin/lib/wp_utils.ml
```diff
@@ -53,22 +52,38 @@ let clear_mapper : Term.mapper = object
...
 end
 
-(* Reads in the program_t and its architecture from a file. *)
-let read_program (ctxt : ctxt) ~(loader : string) ~(filepath : string)
-  : Program.t * Theory.target =
-  let mk_digest = Cache.Digests.get_generator ctxt ~filepath ~loader in
-  let program_digest = Cache.Digests.program mk_digest in
-  match Cache.Program.load program_digest with
-  | Some prog ->
+(* FIXME: read context to determine whether we need the "fat" project or not. *)
```
fortunac, line 55: What is the plan for determining whether a project should be skinny or not?

wp/plugin/lib/wp_utils.ml
```diff
@@ -53,22 +52,38 @@ let clear_mapper : Term.mapper = object
...
+(* Make a "skinny" project with nothing in it *)
+let mk_skinny (proj : project) : project =
+  let tgt = Project.target proj in
+  let empty = Project.empty tgt in
+  let skinny_prog = proj |> Project.program |> clear_mapper#run in
+  let with_prog = Project.with_program empty skinny_prog in
+  with_prog
+
+(* Reads in the project from a file, creating a cache if none exists. *)
+let read_project (ctxt : ctxt) ~(loader : string) ~(filepath : string)
```
fortunac, line 67: Looks good!

wp/lib/bap_wp/src/compare.ml
```diff
@@ -23,6 +23,14 @@ module Env = Environment
 module Pre = Precondition
 module Constr = Constraint
 
+type 'a code =
```
fortunac, line 26: Is this record only used in comparative analyses? Is there a use-case for a single analysis?

wp/lib/bap_wp/src/compare.ml
```diff
@@ -314,3 +332,16 @@ let compare_subs_mem_eq : comparator * comparator =
...
   postcond, hyps
+
+let compare_subs_mem_init : comparator * comparator =
+  let postcond ~original ~modified ~rename_set:_ =
+    let ctx = Env.get_context original.env in
+    let post = Constr.trivial ctx in
+    post, original.env, modified.env
+  in
+  let hyps ~original ~modified ~rename_set:_ =
+    let pre, env1 = Pre.init_mem original.env original.mem in
```
fortunac, line 343: Can you explain why we only need to initialize the mem for the original binary?

wp/plugin/lib/wp_cache.ml
```diff
@@ -32,87 +30,14 @@ module Digests = struct
 
   (* Creates a digest for the knowledge cache. *)
   let knowledge (mk_digest : namespace:string -> digest) : digest =
-    mk_digest ~namespace:"knowledge"
+    mk_digest ~namespace:"wp-knowledge"
```
fortunac, line 33: Were there namespace clashes here?

codyroux: no, actually, It was a hail Mary to try to understand the bug I'm getting.

wp/lib/bap_wp/src/compare.ml
```diff
@@ -140,35 +153,40 @@ let fold_comparators
...
-      let vars, env1, env2 = init_vars env1 env2 vars in
+    let init_vars ~original:code1 ~modified:code2 ~rename_set:vars =
+      let vars, env1, env2 = init_vars code1.env code2.env vars in
       Constr.mk_clause [] vars, env1, env2
     in
-    fold_comparators (init_vars :: hyps) ~original:(sub1, env1)
-      ~modified:(sub2, env2) ~rename_set:vars
+    fold_comparators (init_vars :: hyps)
+      ~original:{ code1 with env=env1 }
+      ~modified:{code2 with env=env2 }
```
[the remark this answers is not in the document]
codyroux: Yep. I hate it when it's inconsistent as well.

wp/lib/bap_wp/src/compare.ml
codyroux: Well, I'm thinking this will mostly apply when the two memories are equal. Maybe I can explicitly introduce a dependency between the flags.

Suggestions are welcome.

wp/plugin/lib/wp_utils.ml
codyroux: It's whether or not the `init-mem` flag is set.

wp/plugin/lib/wp_utils.ml
codyroux: Thanks. currently doesn't work though...

wp/plugin/lib/wp_utils.ml
codyroux: Oh, yea. Also the info string.

wp/lib/bap_wp/src/compare.ml
fortunac: So if the two binaries have the same memories, but at different offsets, then this hypothesis would not apply, right?

codyroux: Indeed. We need to be careful here: if we initialize `mem_mod`, and it's different than `mem_orig` and then assume that they are equal, we'll get an inconsistent assumption, and get spurious `UNSAT`s.

I decided to leave out all assumptions involving `mem_mod` for starters.

wp/lib/bap_wp/src/compare.ml
fortunac: Sorry, my confusion. This was placed in the `compare` module so I was under the assumption that we only used it for comparisons and not single analyses. It felt a bit weird to see some parts of `single` use a type from the `compare` module like

```
| Single of {
      pre : Constr.t;
      orig : sub Comp.code
    }
```

PR #337, opened 2021-08-11
WIP: Change the runner api to expect a project

codyroux:
Also not ready, creating so we have a place for discussion, and a place to point @ivg if he's so inclined.


codyroux:
Currently this is broken: we cannot serialize a Project.

Per @ivg the solution is as follows: save the KB state after the load, and restore it if available. this should greatly speed up project load times.

The relevant code is here: https://github.com/BinaryAnalysisPlatform/bap/blob/master/plugins/disassemble/disassemble_main.ml#L372-L378

PR #338, opened 2021-08-12
Unrolled libc functions

nickroess:
Not ready for merge, here for discussing approach with @codyroux @ccasin


wp/plugin/api/c/cbat_libc.h
```diff
@@ -0,0 +1,252 @@
...
+#endif
+
+#define strcmp cbat_libc_strcmp
+#define strlen cbat_libc_strlen
+#define strcpy cbat_libc_strcpy
+#define strncpy cbat_libc_strncpy
+#define memcmp cbat_libc_memcmp
+#define memcpy cbat_libc_memcpy
+
+#define UNROLL_EXCEEDED __VERIFIER_assume(true)
```
codyroux, line 16: Pretty sure we want this to be `__VERIFIER_assume(false)`, since `assume(true)` is a no-op :)

nickroess: Yeah, that's what I started with but it did not behave as expected. I'll show you on examples when we discuss next

codyroux:
@nickroess Feel free to merge this if there's nothing outstanding.

PR #343, opened 2021-09-10
Add support for multiple user-defined specs

nickroess:
This PR changes `user_func_spec` from an option to a list to allow multiple user specs to be provided at once. Not ready for merge yet, I'm going to add a test


wp/plugin/wp.ml
```diff
@@ -218,7 +218,7 @@ let func_name_map = Cmd.parameter Typ.(list ~sep:';' (pair ~sep:',' string strin
            "<reg1_orig>,<reg1_mod>;<reg2_orig>,<reg2_mod>"). By default, WP
            assumes subroutines have the same names between the two binaries.|}
 
-let user_func_spec = Cmd.parameter Typ.(some (t3 string string string)) "user-func-spec"
+let user_func_spec = Cmd.parameter Typ.(list ~sep:';' (t3 string string string)) "user-func-spec"
     ~doc:{|Creates the weakest precondition for a subroutine given the
```
fortunac, line 222: Can you add how to write multiple user-defined specs and and an example?

wp/plugin/wp.ml
```diff
@@ -218,7 +218,7 @@ let func_name_map = Cmd.parameter Typ.(list ~sep:';' (pair ~sep:',' string strin
            "<reg1_orig>,<reg1_mod>;<reg2_orig>,<reg2_mod>"). By default, WP
            assumes subroutines have the same names between the two binaries.|}
 
-let user_func_spec = Cmd.parameter Typ.(some (t3 string string string)) "user-func-spec"
+let user_func_spec = Cmd.parameter Typ.(list ~sep:';' (t3 string string string)) "user-func-spec"
```
fortunac, line 221: Should it still be `user-func-spec` or should it be changed to `user-func-specs` (with the `s` at the end, or does it not matter)? Does anyone have thoughts?

wp/plugin/wp.ml
fortunac: I also realized that we don't have that flag documented on our site. Can you add it here? https://github.com/draperlaboratory/cbat_tools/blob/master/docs/reference.html

fortunac:
Looks good so far! Can't wait to see the tests.

wp/plugin/wp.ml
codyroux: I have thoughts: it does matter (but not a lot!), and it should be `user_func_specs`.

codyroux: or `user-func-specs` rather, though I'd probably rename the OCaml as well.

nickroess:
@codyroux @fortunac Okay, just updated to
1) Rename `user-func-spec` to `user-func-specs` including all the examples that use the flag
2) Add test for multiple specs that needs 2 specs to get `UNSAT`
3) Update docs with example of multiple specs and added `user-func-specs` to `reference.html`

wp/resources/sample_binaries/user_func_spec/sub_spec_5/run_wp_1.sh
```diff
@@ -0,0 +1,13 @@
+# Should return SAT: we provide a spec only for g, which means f can still return 0x67 and we can trip the assert
+
+run () {
+  bap wp \
+    --func=main \
+    --precond="(assert (= (bvand RDI #xFFFFFFFF00000000) #x0000000000000000))" \
```
codyroux, line 6: Do you need these?

nickroess: Nope! I copied from other tests. Could be removed from this one and others

codyroux: Probably good not to have useless annotations in tests.

nickroess: K they are removed

PR #348, opened 2021-11-11
WIP: user func spec compare

ccasin:


codyroux:
👍

ccasin:
@codyroux did you get UNSAT on the VIBES wp-debug example with this?  I get SAT, which seems wrong, but haven't dug in yet.

PR #349, opened 2021-11-22
Different func specs

ccasin:
This adds two new flags, `--user-func-specs-orig` and `--user-func-specs-mod`.

These can be use to add function specs for one binary or the other in comparative analysis, and to give a function a different spec in each program.  This is based on Cody's earlier branch.  I've changed it to be compatible with the existing `--user-func-specs` flag, added some test cases, and updated the docs.


wp/resources/sample_binaries/user_func_spec/sub_spec_6/run_wp_6.sh
```diff
@@ -0,0 +1,22 @@
...
+
+run () {
+  bap wp \
+    --func=main \
+    --postcond="(assert (and (= RAX_orig (_ bv14 64)) (= RAX_mod (_ bv14 64))))"  \
+    --user-func-specs-orig="g,(assert true),
+(declare-fun g_spec ((_ BitVec 64)) (_ BitVec 64))
+(assert (= RAX (g_spec init_RDI)))" \
+    --user-func-specs-mod="g,(assert true),
+(declare-fun g_spec ((_ BitVec 64)) (_ BitVec 64))
```
fortunac, line 16: How does WP handle the case where the original and modified function specs contain function declarations of the same name?

ccasin: You end up with multiple declarations of this function in the smtlib output, but it works out the way we want: no matter how many times you declare a function, there is just one symbol and all pieces of the formula that mention it are referring to that one symbol.

This was also confusing to me, at first - Cody had to explain it to me twice.  Unfortunately it's necessary to add the declaration in both places because the specs are parsed individually, so you get an "unknown constant `g_spec`" error if you leave either off.

In the long term, it would be nice to add a flag to CBAT where the user could specify some kind of "prelude" that is automatically pasted before any other smtlib strings they provide.

fortunac: So both sides need to have the function declaration if its name is being used as an assertion? Do you think it's worth having this note in the docs for these flags? What happens if the modified spec changes the sort of the symbol?

ccasin: Yes - these pieces of smtlib are parsed independently, so any new names the user wants to refer to need to be declared in each.

I think this use case is probably not the first thing people would do with these flags (see `run_wp_1.sh` for an example with no new declarations), so it probably doesn't make sense to really go into detail in the description, but noting the above fact is reasonable.

I believe that in smtlib you are allowed to use the same name for two distinct symbols with different types (and I just tested very quickly).  So, if you change the type, you now have two unrelated functions.

Just to be doubly clear about what's going on here:  Imagine your program calls some function `g`.  Perhaps it's a complicated function that you don't want to specify in great detail.  But you do know that the modified program has changed `g` slightly (e.g., perhaps it first adds 1 to its input and then does the same thing as the old `g`).  We want to specify this relationship between the functions, so that `wp` can see calls to `g(n)` in the modified binary are the same as calls to `g(n+1)` in the original binary.

How do we do this without specifying the complete behavior of `g`?  That's the trick here: we declare a new uninterpreted function called `g_spec`, and we say that `g(n)` in the original program does whatever `g_spec(n)` does.  Then we say that `g(n)` in the modified program has the behavior `g_spec(n+1)`.  We haven't said anything about what `g_spec` does, but we've established a relationship between the two versions of `g` using it.

The fact that we need to declare `g_spec` in both places is just an artifact of the fact that these user function summaries are parsed as little independent bits of smtlib.  I don't see an easy fix, but if it helps you can think about it like C - where multiple files can declare the same function whose implementation is provided in some other place.

ccasin: (The above use case it what is tested in `run_wp_2.sh`.  This file, `6`, is an example showing we don't actually know anything specific about the behavior of `g_spec`)

PR #350, opened 2021-11-28
Ccasin/user func spec modified vars fix

ccasin:
Changes how we calculate what system state is modified when using a user function spec.

The old story was that we analyzed the actual implementation of the function to see what state it touched, and treated everything it touched as both an input an an output to the function.  This had a few issues.  Most pressingly, it doesn't work at all for functions where the implementation is unavailable or can't be analyzed.  It also meant we overestimated what state is changed by a function.

The new system is to look at the variables in the user-provided function spec.  Uses of variables corresponding to the initial values of registers/memory mean those are inputs, and uses of variables corresponding to the final values of registers/memory mean those are updated by the function.  This works well enough - the tests pass (I haven't added new tests since it is exercised by all our existing user function spec tests).

### DIFF
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -1303,7 +1303,7 @@ let user_func_spec
          kind of empty constr. *)
 
       (* Next, collect inputs/outputs of sub.  We do this by constructing a set
-         of all potential inputs/outputs (all the regsiters and mem), then
+         of all potential inputs/outputs (all the registers and mem), then
          filtering out everything not mentioned by the user-provided specs.
 
          Question: is SP + GPRs really all the registers?  *)
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -1304,12 +1304,8 @@ let user_func_spec
 
       (* Next, collect inputs/outputs of sub.  We do this by constructing a set
          of all potential inputs/outputs (all the registers and mem), then
-         filtering out everything not mentioned by the user-provided specs.
-
-         Question: is SP + GPRs really all the registers?  *)
-      let potential_ios : Bap.Std.Var.Set.t =
-        Set.add (Set.add (Env.get_gprs env) (Env.get_sp env)) (Env.get_mem env)
-      in
+         filtering out everything not mentioned by the user-provided specs. *)
+      let potential_ios : Bap.Std.Var.Set.t = Env.get_all_target_vars env in
       let filter_out_unused (used : string list) (adjust : string -> string) =
         Set.filter potential_ios
           ~f:(fun v ->
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -1289,31 +1289,24 @@ let user_func_spec
   debug "Making user-defined subroutine spec with subroutine-name: %s, pre:
 %s, post: %s \n%!" sub_name sub_pre sub_post;
   if String.equal sub_name (Sub.name sub) then
-    let spec env post _tid = (
+    let spec env post _ = (
       (* turn strings into proper smtlib2 statements; incr stack_ptr *)
       let (sub_pre, pre_init_vars, pre_vars)
           : Constr.t  * string list * string list =
-        Z3_utils.mk_smtlib2_single_with_vars ~debug:true env sub_pre in
+        Z3_utils.mk_smtlib2_single_with_vars env sub_pre in
       let (sub_post, post_init_vars, post_vars)
           : Constr.t * string list * string list =
-        Z3_utils.mk_smtlib2_single_with_vars ~debug:true env sub_post in
-      (* TODO: Some sanity checking on sub_pre and sub_post here.  If you
-         just forget the word "assert" you silently get some kind of empty
-         constr. *)
+        Z3_utils.mk_smtlib2_single_with_vars env sub_post in
       let sub_post, env = increment_stack_ptr sub_post env in
+      (* TODO: Some sanity checking on sub_pre and sub_post would be nice
+         here.  If you just forget the word "assert" you silently get some
+         kind of empty constr. *)
 
-      (* collect (physical) inputs/outputs of sub *)
-      let sub_inputs : Var.t list = (vars_from_sub env sub |> Var.Set.to_list) in
-      let _sub_inputs =
-        List.filter sub_inputs ~f:(fun v -> Var.is_physical v) in
-      let _sub_outputs : Var.t list = sub_inputs in
-
-      (* collect (physical) inputs/outputs of sub.  We do this by constructing a
-         set of all potential inputs/outputs (all the regsiters and mem), then
+      (* Next, collect inputs/outputs of sub.  We do this by constructing a set
+         of all potential inputs/outputs (all the regsiters and mem), then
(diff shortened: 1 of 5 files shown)


ccasin:
For the implementation, there are two hurdles: 1) get the variables, as strings, from the user-provided spec, and 2) find the corresponding `Var.t`s (since that is what is needed by the function summary code).

For 1: The user specs are parsed by `mk_smtlib2_single`.  This function already finds the variables we care about, so it seems easiest to just have it return them in addition to the constraint.  So I made a new version of it, `mk_smtlib2_single_with_vars` that does.  (I changed `mk_smtlib2_single` to be defined in terms of it, to avoid code duplication).

For 2: It wasn't immediately obvious to me the best way to go from the variable `string`s to the corresponding `Var.t`s.  I ended up just getting the set of `Var.t`s for all registers and memory from the `env`, and then filtering out anything that's not mentioned in the user spec.

Let me know if there's a better way to do either of these.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1289,16 +1289,39 @@ let user_func_spec
...
+      (* TODO: Some sanity checking on sub_pre and sub_post would be nice
+         here.  If you just forget the word "assert" you silently get some
+         kind of empty constr. *)
+
+      (* Next, collect inputs/outputs of sub.  We do this by constructing a set
+         of all potential inputs/outputs (all the registers and mem), then
+         filtering out everything not mentioned by the user-provided specs.
+
+         Question: is SP + GPRs really all the registers?  *)
+      let potential_ios : Bap.Std.Var.Set.t =
```
fortunac, line 1310: This was the reason why we looked at the subroutine, and found the union of SP, the GPRs, and the vars found in the sub just in case we were missing registers. It seems we can't do that here since we won't always have the body of the subroutine. I wonder if http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/Target/index.html#val-vars gives us something different.

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1289,16 +1289,39 @@ let user_func_spec
...
+      in
+      let filter_out_unused (used : string list) (adjust : string -> string) =
+        Set.filter potential_ios
+          ~f:(fun v ->
+            List.exists used
+              ~f:(fun s -> String.equal (adjust (Var.to_string v)) s))
+      in
+      let (sub_inputs, sub_outputs : Var.t list * Var.t list) =
+        (Set.to_list (filter_out_unused (pre_init_vars @ post_init_vars)
+                        (fun s -> "init_" ^ s)),
```
fortunac, line 1321: We might want to clean the name before adding the `init_` like we do over here: https://github.com/draperlaboratory/cbat_tools/blob/7bb7e36b247860d81f8b4ce43779c1ceb1b76547/wp/lib/bap_wp/src/environment.ml#L443

wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1307,7 +1330,7 @@ let user_func_spec
           match r with
           | Some q -> q
           | None -> let q, _ = Env.mk_init_var env v in q) in
-      let tid_name : string = Tid.name tid in
+      let tid_name : string = Tid.create () |> Tid.to_string in
```
fortunac, line 1333: Why do we create a new tid here?

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -132,27 +132,38 @@ let mk_smtlib2_compare ?(name = None) (env1 : Env.t) (env2 : Env.t)
...
+    List.fold_right tokens ~init:([],[],[]) ~f:(fun t (ts,pre_v,post_v) ->
+        match get_z3_name var_map t Var.name with
+        | Some n -> (Expr.to_string n :: ts, pre_v, t :: post_v)
         | None ->
           begin
-            match get_z3_name init_var_map token (fun v -> "init_" ^ Var.name v) with
-            | Some n -> n |> Expr.to_string
-            | None -> token
+            match get_z3_name init_var_map t
+                    (fun v -> "init_" ^ Var.name v) with
```
fortunac, line 152: We might want to clean the name here too.

fortunac:
> For the implementation, there are two hurdles: 1) get the variables, as strings, from the user-provided spec, and 2) find the corresponding `Var.t`s (since that is what is needed by the function summary code).
> 
> For 1: The user specs are parsed by `mk_smtlib2_single`. This function already finds the variables we care about, so it seems easiest to just have it return them in addition to the constraint. So I made a new version of it, `mk_smtlib2_single_with_vars` that does. (I changed `mk_smtlib2_single` to be defined in terms of it, to avoid code duplication).
> 
> For 2: It wasn't immediately obvious to me the best way to go from the variable `string`s to the corresponding `Var.t`s. I ended up just getting the set of `Var.t`s for all registers and memory from the `env`, and then filtering out anything that's not mentioned in the user spec.
> 
> Let me know if there's a better way to do either of these.

I can't think of a better solution for those issues. One pressing question I have would be if there is ever a case where the user has a virtual variable in their spec. If there is a case, are we able to handle this?

wp/lib/bap_wp/src/precondition.ml
ccasin: Good idea.  I tested this and it returns everything we were already getting, plus some flag registers.  So I added a function to `environment.ml` like `get_gprs` and used it here.

wp/lib/bap_wp/src/precondition.ml
ccasin: I think this is not needed.  We want to check whether the user has written _exactly_ one of the register/memory names. right?  Like `RAX` or `init_RAX`.  If we cleaned the name wouldn't we risk mistaking something else the user wrote for a register name or memory?

Perhaps I'm missing some subtlety though.

wp/lib/bap_wp/src/precondition.ml
ccasin: The old behavior was a bug in the case where a summarized function is called multiple times.  It resulted us generating the same name for the return memories of the different calls.

wp/lib/bap_wp/src/z3_utils.ml
ccasin: I guess I don't totally understand what name cleaning does for us.  It looks like the clean function just drops the `|` character.  I'm guessing that's because z3 doesn't like it?  But for the case of a user-provided specification, if they've used this character wouldn't it be better to have z3 fail loudly than to silently assume, for example, they meant `RAX` instead of `RA|X`?

ccasin:
Your last question is a good one, but I think we're fine.  The way `sub_inputs` and `sub_outputs` are calculated is that we start from the set of register/memory names we got out of the environment, and we then keep only the ones that match a name in the user spec.  We are never adding random things from the user spec to this set.

wp/lib/bap_wp/src/z3_utils.ml
fortunac: Makes sense. We mostly clean those so Boolector doesn't complain. The case I was thinking about was when a user has a virtual variable in their spec, but your last comment convinced me otherwise. If a user specified a virtual variable in their spec, the variable they wrote wouldn't match what is already in the environment since we clean those names there. If there are no virtual variables, then this is no problem, and we should keep it like this.

PR #362, opened 2022-03-24
Add support for using the BAP ogre lifter via a wp command line option

ccasin:
This lets us use ogre files to tell the BAP lifter to lift a specific portion of the binary.  It has two main advantages: 1) We can lift only the part we care about from big binaries, saving time.  2) We can use the ogre file to give names to functions in stripped binaries.


wp/plugin/tests/unit/test_wp_unit.ml
```diff
@@ -151,6 +151,19 @@ let unit_tests = [
   "User defined sub specs comparative change 6" >: test_plugin "user_func_spec/sub_spec_6"
     sat ~script:"run_wp_6.sh";
 
+  "Ogre lifter 1" >: test_plugin "ogre_files/ogre_lifter_1"
```
fortunac, line 154: I think the path should be changed to `ogre_files` since `ogre_lifter_x` files don't exist.

Same with `run_wp_x.sh` should be changed to `run_wpx.sh`.

Issue #366, opened 2022-06-10
Incorrect detection of intrinsic functions when there is no BIL lifter available


bmourad01:
On VIBES we're working mainly with Thumb binaries, and in BAP a substantial portion of the semantics for Thumb instructions are not being provided by any of the available BIL lifters. However, the current way that calls to intrinsic functions is being detected relies on there being a BIL lifter available.

As a consequence, we're unable to provide any meaningful info (such as user function specs) about function calls in these binaries. We should change how such intrinsic calls are detected.

bmourad01:
Fixed in #367

PR #367, opened 2022-06-16
Fixes detection of intrinsic calls in BAP

bmourad01:
Addresses #366.

We no longer rely on a BIL lifter being available, but instead look at
the subroutine in question since the information is there on whether it
is intrinsic or not.

The `Sub.intrinsic` tag doesn't seem to get populated anywhere, so for
now we can check that the subroutine's name is prefixed with
`intrinsic:`.


fortunac:
The tests in the CI are failing. Some of them return the opposite result from what we expect while others return errors. Do you know what's causing this?

bmourad01:
I think it's what we discussed related to the binaries not being lifted correctly. None of these tests are related to detecting intrinsic calls though.

PR #368, opened 2022-06-24
Changes `init-mem` to look at all read-only segments

bmourad01:
The `--init-mem` option currently tells WP to initialize everything in the `.rodata` section for both the original and modified binaries. In some binaries, we might be interested in code that gets inlined into the `.text` section (e.g. constant pools on ARM/Thumb).

This PR changes the behavior to look at any segment without write permissions. Since this will include the executable segments, it can potentially cause a blowup in memory usage. To fix this, we cache the set of known code addresses by calling `Disasm.Driver.explore` on entry to the library/plugin.

On a side note, if an OGRE file was provided, then `Utils.init_mem` should use that as the loader instead of the default loader (currently LLVM). This can save a lot on memory usage and running time.

### DIFF
--- a/wp/lib/bap_wp/src/precondition.ml
+++ b/wp/lib/bap_wp/src/precondition.ml
@@ -1282,9 +1282,9 @@ let init_vars (vars : Var.Set.t) (env : Env.t) : Constr.t list * Env.t =
    want to initialize. *)
 let init_mem_section (mem : value memmap) : (value memmap) =
   mem |> Memmap.filter ~f:(fun v ->
-      match Value.get Image.section v with
+      match Value.get Image.segment v with
       | None -> false
-      | Some name -> String.equal name ".rodata")
+      | Some seg -> not @@ Image.Segment.is_writable seg)
 
 let init_mem_range (ctx : Z3.context) (init_mem : bool)
     (mem_orig : value memmap) (mem_mod : value memmap) (addr :  Constr.z3_expr)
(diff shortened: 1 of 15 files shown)


wp/lib/bap_wp/src/precondition.ml
```diff
@@ -1310,28 +1310,33 @@ let init_mem_range (ctx : Z3.context) (init_mem : bool)
     (* The mem read does not take place in the initialized section. *)
     Bool.mk_false ctx
 
-let init_mem (env : Env.t) (mem : value memmap) : Constr.t list * Env.t =
+let init_mem ?(code_addrs : Utils.Code_addrs.t = Utils.Code_addrs.empty)
```
fortunac, line 1313: Do we ever anticipate `code_addrs` to be empty in practice? Or is it just for the test suite?

bmourad01: Mainly for the test suite. It currently gets computed and cached no matter what, but an idea would be to not do this if the `init-mem` parameter is disabled.

fortunac: I think `Pre.init_mem` doesn't get called when the `init_mem` parameter is disabled. In the test suite, I noticed that `Code_addrs.empty` is provided. Is it necessary to keep this argument optional?

fortunac: That's a fair point. How does the cache behave if the code addrs aren't initially included in the cache and the user runs WP again with the `init_mem` flag on?

PR #370, opened 2022-08-29
Clarify that different versions/installations of CBAT might produce different TIDs and solutions, and fix small typos.

slasser:


jtpaasch (approved):
Thank you Sam!

PR #371, opened 2022-09-16
Create flag for turning off the chaos spec

fortunac:
This PR creates the `--no-chaos=<function list>` flag. Normally, functions like malloc would use the `verifier_nondet` function spec. With this flag set, the specified functions will not use the `verifier_nondet` spec, but will be treated as all other functions. For each function, it will look at the list of default function specs, and apply the spec who's requirements are satisfied.


jtpaasch (approved):
👍

PR #372, opened 2022-10-06
Jt/limit opam jobs

jtpaasch:
* Dockerfile allows `OPAM_JOBS` build arg to limit opam from taking over your 'puter.
* Installation documentation updated.
* Developer API docs regenerated.


Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM ocaml/opam:ubuntu-20.04-ocaml-4.14
 
+ARG OPAM_JOBS=0
```
fortunac, line 3: Is this something we want the user to choose with a build-arg or just use `nproc`?

jtpaasch: Hmmm, that's not a bad idea. Let me experiment with that.

fortunac:
Looks good! Merge at will.

fortunac (approved):

jtpaasch:
Okay, one more little thing Chloe. I removed that problematic `opam install .` line, so that we don't break for people running opam 2.1, and I added a BRANCH build-arg to the dockerfile so that one can build the docker container from a custom branch.

Sound okay to you?

fortunac:
> Okay, one more little thing Chloe. I removed that problematic `opam install .` line, so that we don't break for people running opam 2.1, and I added a BRANCH build-arg to the dockerfile so that one can build the docker container from a custom branch.
> 
> Sound okay to you?

Go for it!

PR #373, opened 2022-11-03
Add diagnostics to show flag

fortunac:
Adds the flag for printing diagnostics. If the flag is set, the "evaluating constraint" message, the "sending to Z3" message, the output of "SAT/UNSAT/UNKNOWN", and the model is sent to stderr.


jtpaasch (approved):
👍

PR #374, opened 2022-11-03
Add exit codes to wp

fortunac:
Adds exit codes to WP:
unsat -> 0
sat -> 1
unknown -> 2

### DIFF
--- a/wp/plugin/wp.ml
+++ b/wp/plugin/wp.ml
@@ -431,8 +431,10 @@ let callback
     })
   in
   Parameters.validate params files >>= fun params ->
-  Analysis.run params files ctxt >>= fun () ->
-  Ok ()
+  Analysis.run params files ctxt >>= function
+  | Z3.Solver.UNSATISFIABLE -> exit 0
+  | Z3.Solver.SATISFIABLE -> exit 1
+  | Z3.Solver.UNKNOWN -> exit 2
 
 let () =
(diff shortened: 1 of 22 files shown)


wp/plugin/wp.ml
```diff
@@ -431,8 +431,10 @@ let callback
     })
   in
   Parameters.validate params files >>= fun params ->
-  Analysis.run params files ctxt >>= fun () ->
-  Ok ()
+  Analysis.run params files ctxt >>= function
+  | Z3.Solver.UNSATISFIABLE -> exit 0
```
jtpaasch, line 435: Hmm...I don't know if BAP needs to do any special cleanup after a custom command returns. If so, maybe it's not a good idea to force a premature exit here.

I wonder if there's a way to return `Error ...` with specific exit codes. Maybe a question for Ivan?

Or, maybe we could return `Ok ()` for UNSAT, `Error ...` for SAT, and then `exit 2` for UNKNOWN?

jtpaasch (approved):
Marking this approved, but there's a comment/question. Do with it what you will!

wp/plugin/wp.ml
fortunac: That's a very good point. I'll take a look into it!

jtpaasch:
Fantastic work on fixing this.

PR #378, opened 2023-01-10
Builds the plugin using dune

bmourad01:
Since BAP switched the default build system to dune, we should do the same for our plugins.


wp/plugin/Makefile
```diff
@@ -97,15 +65,8 @@ test: test.unit
 
 .PHONY: test.unit
 test.unit: install
-	ocamlbuild -r -use-ocamlfind \
-		-pkgs $(TEST_PKGS) \
-		-tag $(TEST_TAG) -Is $(TEST_Is_UNIT) test.native
-	./test.native
+	dune runtest tests/unit
```
ivg, line 68: I suggest adding `--force` here, so that if bap is changed the tests are still be running (the default behavior of dune is to skip test if their dependencies wasn't changed)

wp/plugin/wp.opam
```diff
@@ -0,0 +1,25 @@
+opam-version: "2.0"
```
ivg, line 1: This opam file misses the build and install steps, which is a bit surprising. I would suggest using dune-project to encode your dependencies. Not only it has nicer syntax, but it also properly handles plugin installations. You can use the wp.opam.template file if you need to override any fields.

wp/plugin/wp.opam
```diff
@@ -0,0 +1,25 @@
...
+  "JT Paasch <jpaasch@draper.com>"
+  "Philip Zucker <pzucker@draper.com>"
+  "Benjamin Mourad <bmourad@draper.com>"
+]
+license: "MIT"
+homepage: "https://github.com/draperlaboratory/cbat_tools"
+bug-reports: "https://github.com/draperlaboratory/cbat_tools"
+dev-repo: "git+https://github.com/draperlaboratory/cbat_tools"
+depends: [
+  "bap" {>= "2.4.0"}
```
ivg, line 20: You need here dune-powered bap, which is `bap {>= "2.6.0"}`

ivg (approved):
Some minor suggestions, otherwise looks good!

XVilka:
There are conflicts now. Any updates to get this merged, by the way?

PR #380, opened 2023-02-24
Fixes not extracting the memory model from lambda terms

bmourad01:
With some external solvers (like Boolector), the memory for the countermodels comes back as a big lambda term consisting of a bunch of if-then-else cases for each address.

Obviously, this is very unreadable for the user. Not only that, WP will fail to construct the refuted goals when the memory model is in this form. 

This PR fixes the way we parse the output from the external solver, turning it into a sequence of array stores.

### DIFF
--- a/wp/lib/bap_wp/src/output.ml
+++ b/wp/lib/bap_wp/src/output.ml
@@ -28,23 +28,33 @@ module FInterp = Model.FuncInterp
 module Env = Environment
 module Constr = Constraint
 module Comp = Compare
+module Quant = Z3.Quantifier
+module BV = Z3.BitVector
+module Sym = Z3.Symbol
+module AST = Z3.AST
 
 module VarMap = Var.Map
 
-type mem_model = {default : Constr.z3_expr ; model : (Constr.z3_expr * Constr.z3_expr) list}
+type mem_model = {
+  default : Constr.z3_expr;
+  model : (Constr.z3_expr * Constr.z3_expr) list
+}
+
+let equal_mem_model (m1 : mem_model) (m2 : mem_model) : bool =
+  let eqe (k1, v1) (k2, v2) = Expr.equal k1 k2 && Expr.equal v1 v2 in
+  let eqd = Expr.equal m1.default m2.default in
+  let eqm = List.equal eqe m1.model m2.model in
+  eqd && eqm
 
 let format_mem_model (fmt : Format.formatter) (mem_model : mem_model) : unit =
-  mem_model.model
-  |> List.sort ~compare:(fun (addr1, _) (addr2, _) ->
-      String.compare (Expr.to_string addr1) (Expr.to_string addr2))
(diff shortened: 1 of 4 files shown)


wp/lib/bap_wp/src/output.ml
```diff
@@ -43,39 +44,87 @@ let format_mem_model (fmt : Format.formatter) (mem_model : mem_model) : unit =
   then Format.fprintf fmt "\t\telse |-> %s]\n" (Constr.expr_to_hex mem_model.default)
   else Format.fprintf fmt "\t\t%s]\n" (Expr.to_string mem_model.default)
 
-(** [extract_array] takes a z3 expression that is a seqeunce of store and converts it into
-    a mem_model, which consists of a key/value association list and a default value *)
-
+(* Takes a z3 expression that is either a seqeunce of stores or a lambda term
```
fortunac, line 47: Super small nitpick, can you fix the typo in sequence?

wp/lib/bap_wp/src/output.ml
```diff
@@ -27,55 +28,116 @@ module FInterp = Model.FuncInterp
...
+  let eqd = Expr.equal m1.default m2.default in
+  let eqm = List.equal eqe m1.model m2.model in
+  eqd && eqm
 
 let format_mem_model (fmt : Format.formatter) (mem_model : mem_model) : unit =
-  mem_model.model
-  |> List.sort ~compare:(fun (addr1, _) (addr2, _) ->
-      String.compare (Expr.to_string addr1) (Expr.to_string addr2))
-  |> List.iter ~f:(fun (key, data) ->
+  mem_model.model |> List.iter ~f:(fun (key, data) ->
```
fortunac, line 50: Is this list always going to be in order by address?

bmourad01: As long as it's the one being produced by `extract_array`, then yes. I think that this function `format_mem_model` is not used anywhere outside of this module.

wp/lib/bap_wp/src/z3_utils.ml
```diff
@@ -169,12 +177,35 @@ let mk_smtlib2_single ?(name = None) (env : Env.t) (smt_post : string)
...
     [and] node if the number of operands is less than 2. This may improve sharing,
     but also improves compatibility of smtlib2 expressions with other solvers  *)
-let mk_and ( ctx : Z3.context ) (xs : Constr.z3_expr list) : Constr.z3_expr =
+let mk_and (ctx : Z3.context) (xs : Constr.z3_expr list) : Constr.z3_expr =
   match xs with
   | []  -> Bool.mk_true ctx
   | [x] -> x
   | _   -> Bool.mk_and ctx xs
 
+let process_lambda
```
fortunac, line 186: Can we add a comment about what we are processing lambdas into?

fortunac (approved):

PR #382, opened 2023-02-24
pycbat MVP

philzook58:
The idea is to make cbat easier to use. A minimal (but still kind of beefy. 80mb compressed. 300mb uncompressed) docker build is now up and this python is a wrapper ~to download and~ run it.

Todo:
- Fill out and forward more command line options
- Eventually upload this to pip for `pip install cbat`
- Property construction helpers / connection to Highvars. C type interpretations. Signature importing. Suggest using angr.
- Countermodel exploration, visualization, and interpretation to highvars. Perhaps run in angr symbolic executor
- Port tutorial to jupyter notebook hostable on Google colab links


wp/pycbat/README.md
```diff
@@ -0,0 +1,23 @@
...
+Usage:
+
+```python
+import z3
+import cbat
+
+rax = z3.BitVec("RAX", 64)
+postcond = rax == 4
+# Check that rax is always 4 at end of function "fact" in binary file myfactorial.o
+cbat.run_wp("./myfactorial.o", func="fact", postcond=postcond)
```
fortunac, line 21: Can we include more documentation about the arguments supported by `run_wp`?

wp/pycbat/requirements.txt
```diff
@@ -0,0 +1,3 @@
+z3
```
fortunac, line 1: Should we version lock these?

